PR #11, opened 2016-04-15
Generic ToDBTask

shyamd:
Basic idea of ToDBTask that would allow for 'any' Drone to operate. 

Note: Fixed serialization with MMVaspToDbTaskDrone. The drone in pymatgen-db breaks the serializer by not writing @class and @module. Doesn't seem to hurt current operations as we never serialize a drone in our workflows for now.


computron:
Thanks!  A couple of initial notes:

1. Note that MatMethods uses a line length of 100 rather than the PEP default of 79. I think the PEP guideline makes code less readable more often than helps. So no need to add line breaks at ~79 chars. Note that I am not the only one that thinks the PEP guideline is mistaken on this.

2. I can see why you prefer the more general "calc_loc" name rather than "vasp_loc". But I don't think there is a point in having both given that (as far as I can tell) they are the exact same thing. Can you just make it all be "calc_loc" everywhere and remove all references to VaspLoc and vasp_loc?

3. Please add your name to files that you are contributing to

computron:
Also it looks like the branch has conflicts - can you merge them?

shyamd:
1.) Yup sounds good. Didn't know you weren't going by that guidelines. 
2.) Done
3.) Done

computron:
Thanks!

Issue #12, opened 2016-04-18
issues in drone parsing


computron:
@matk86 I am looking over the results of the MatMethods drone parsing a run with both "relax1" and "relax2" jobs and see several issues:

1. calcs_reversed.0.completed_at is *earlier* than calcs_reversed.1.completed_at. Remember that the zero index is supposed to be the later job. Do you know why this is? Other properties might be OK and in the right order. e.g. I checked the final energies and the zero index indeed gives the correct final energy of relax2. But the "completed_at" looks wrong, and maybe other things are too? Can you check this?

2. the "run_stats" (root key) has sub-keys "standard" and "relax2". This should be "relax1" (not standard) and "relax2". Mainly, I feel the sub-keys in "run_stats" should match the labels in "calcs_reversed.x.task.name" if the latter are present.

3. In the root "output" key, there is an "outcar" subkey. I would delete this outcar subkey; it is redundant with calcs_reversed.0.output.outcar. Besides, it looks like it is using the *wrong* outcar (the calcs_reversed.1.output.outcar)!!!

You should be able to reproduce the above errors by running your drone on:
/MatMethods/matmethods/vasp/tests/reference_files/Si_structure_optimization_relax2/outputs

However, in case you are not seeing the above problems, let me know. The actual problems I saw were in a directory at NERSC (/global/project/projectdirs/m2439/matmethods_test/scratch/block_2016-04-15-21-06-54-570920/launcher_2016-04-15-21-06-54-727961) and I will need to give you access.

matk86:
I will look into it

matk86:
1. all the files in the directory: /MatMethods/matmethods/vasp/tests/reference_files/Si_structure_optimization_relax2/outputs, yield same value for str(datetime.datetime.fromtimestamp(os.path.getmtime(vasprun_file))).

2. fixed the runstats taskname issue in the post_process method

3. removed outcar key from the root 'output', moved it to calcs_reversed.x.output.outcar(this was not here before)

the fixes are in the latest pull request

matk86:
please dont merge now, found some more bugs..fixing it

computron:
ok - let me know when finished. I will also send over the directory with the completed_at issue now

matk86:
ok thanks

matk86:
fixed

PR #13, opened 2016-04-19
issue #12 fixes

matk86:
Major overhaul of the drone:

1. subclasses the AbstractDrone directly instead of the pymatgen-db vaspdrone. The drone overloads almost every single method from pymatgen-db drone anyway. 

2. Renamed it to VaspToDbtaskDrone(removed the leading 'MM')

3. added outcar to calc_reversed.x.output.outcar key. The drone filters and assimilites both vasprun.xml and outcar files.

4. post_process method is only for processing non-vasp generated files such as transformation.json and custodian.json

5. added more drone unittests

6. the 'completed_at' key seems to be working fine with the fe2o3 data send by @computron. But if you track the files using git, the time of modification for all the files become identical ==> str(datetime.datetime.fromtimestamp(os.path.getmtime(vasprun_file))) will return same value for *.relax1.gz and *.relax2.gz.

### DIFF
--- a/matmethods/vasp/drones.py
+++ b/matmethods/vasp/drones.py
@@ -1,7 +1,6 @@
 # coding: utf-8
 
-from __future__ import division, print_function, unicode_literals, \
-    absolute_import
+from __future__ import division, print_function, unicode_literals, absolute_import
 
 """
 This Drone tries to produce a more sensible task dictionary than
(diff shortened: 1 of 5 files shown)


matk86:
Major overhaul of the drone:

1.    subclasses the AbstractDrone directly instead of the pymatgen-db vaspdrone. The drone overloads almost every single method from pymatgen-db drone anyway.

2.    Renamed it to VaspToDbtaskDrone(removed the leading 'MM')

3.    added outcar to calc_reversed.x.output.outcar key. The drone filters and assimilites both vasprun.xml and outcar files.

4.    post_process method is only for processing non-vasp generated files such as transformation.json and custodian.json

5.    added more drone unittests

6.    the 'completed_at' key seems to be working fine with the fe2o3 data send by @computron. But if you track the files using git, the time of modification for all the files become identical ==> str(datetime.datetime.fromtimestamp(os.path.getmtime(vasprun_file))) will return same value for *.relax1.gz and *.relax2.gz.

PR #14, opened 2016-04-20
get the bandstructure properties from Bandstructure instead of Vasprun

matk86:
do you have any sample runs where the vaspsrun numbers dont agree with the bandstructure ones? For Si data in the reference files the values match.


computron:
It seems to happen for metals and in particular Al ncsf run. See for example:

/project/projectdirs/m2439/matmethods_test/scratch/block_2016-04-15-21-06-54-570920/launcher_2016-04-15-21-11-19-307436

I just gave you access to m2439 repo on NERSC so hopefully you can see it. Ping me if you can't.

matk86:
says access denied

matk86:
added basic remote filesystem support using paramiko + fix the bandstructure data setting in the drone + unittests for that

computron:
A couple of comments:

1 - The bandstructure patch looks correct functionally, but do you think it is better to change anything upstream in pymatgen? It is not exactly clear to me why in the first place it matters where I am getting the gap from, especially for a clearly metallic compound like Al it seems pymatgen should always give 0.

2 - The paramiko functions look useful, but do they belong in MMethods or some other code? I am happy to leave them in MMethods for the time being, but they seem to be a separate function. e.g. something that MPWorks users might also use.

Let me know what you think before I merge.

matk86:
1. I think the issue with the vasprun vbm and cbm(hence bandgap) is that it assumes both vbm and cbm occurs at the same k point(in essence it is similar to the get_direct_gap in Bandstructure).

2. Since its in the trial phase i would like to leave it in the MM for the time being.

computron:
>> 1. I think the issue with the vasprun vbm and cbm(hence bandgap) is that it assumes both vbm and cbm occurs at the same k point(in essence it is similar to the get_direct_gap in Bandstructure).

If I understand correctly, this sounds like a bug in pymatgen?

2 - OK

matk86:
seems like it, i will have to  read the code more carefully

computron:
I glanced at the PMG code quickly. I didn't see the same problem as you, i.e., it looked like Vasprun could indeed have different k-points for VBM and CBM. However, the BandStructure gap explicitly checks to see "is_metal" (which is a different algorithm that depends on efermi) and sets an exact gap of zero when this is the case. The "is_metal" check could be another reason for the difference

matk86:
yes, you are right.. vasprun doesnt have the is_metal filter that checks for band crossing...

matk86:
by the way i have added "is_metal" to the drone in MM.

computron:
One easy way to make them consistent (especially since it is almost the same algorithm repeated) is to have Vasprun.eigenvalue_band_properties be implemented simply get doing "Vasprun.get_band_structure().get_band_gap()" (which returns VBM, CBM, gap, and direct-ness). Thus the Vasprun.eigenvalue_band_properties simply constructs a BandStructure object and uses it to get the gap stuff.

What do you think?

I am not sure if there is supposed to be any difference between the two values, but it seems strange to me that they are different.

matk86:
sounds good, atleast it will be consistent

computron:
@shyuep - What do you think about making Vasprun.eigenvalue_band_properties simply use get_band_structure().get_band_gap() instead of explicitly implementing the parse of CBM, VBM, etc?

Currently, the two functions can give inconsistent results for metals. e.g. for Al it is 0.0 eV exactly for get_band_structure().get_band_gap() but something like 0.0003 eV for Vasprun.eigenvalue_band_properties.

computron:
Ok what I am going to do is merge your changes for now and create an issue ticket on pymatgen GH page. If pymatgen revises its algorithm for Vasprun I will ask you to revise MM again

matk86:
ok

PR #20, opened 2016-05-24
Md fwork

aykol:
## Summary

Added a simple MD firework. Used MITMDVaspInputSet for now.

## TODO (if any)

MD needs its some specific parsing of outputs. VaspToDbTask should probably be modified.

### DIFF
--- a/matmethods/vasp/firetasks/run_calc.py
+++ b/matmethods/vasp/firetasks/run_calc.py
@@ -112,16 +112,16 @@ class RunVaspCustodian(FireTaskBase):
     def run_task(self, fw_spec):
 
         handler_groups = {
-        "default": [VaspErrorHandler(), MeshSymmetryErrorHandler(),
-                    UnconvergedErrorHandler(), NonConvergingErrorHandler(),
-                    PotimErrorHandler(), PositiveEnergyErrorHandler()],
-        "strict": [VaspErrorHandler(), MeshSymmetryErrorHandler(),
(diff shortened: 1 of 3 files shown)


computron:
Thanks! A few notes:

1. Be careful with - ``env_chk(self["gamma_vasp_cmd"], fw_spec)`` - make sure you use .get() as in the original method. Can you fix?

2. I don't like the negative handlers thing. Way too unintuitive. Since we are still in alpha, feel free to reorganize the handler levels (and change the default value so that it gives the same behavior). If that doesn't work, maybe forget the numerical scale entirely and allow the user to specify a "handler_group" instead of "handler_lvl" - where the group is a string like "md" or "default" or "strict" etc and defines a clear set of handlers. Make sure to define the groups in an easy to see location, e.g. in the header of the Task, where you can define a dictionary of keys and the handlers they define.

aykol:
Thanks for the great feedback! I will fix these and resubmit.

aykol:
All done.

computron:
Thanks! I am merging now. Note that when changing things like handler_lvl --> handler_group, it is always worth doing a search for handler_lvl in all the codebase (and any dependent codebases that know of). In this case, there are a few comments where it is referenced (nothing major, but worth updating - I will do it)

computron:
Nevermind my comment! For some reason I did not see that you already updated the text handler_lvl everywhere. Your changes were fine as-is.

aykol:
Great. Thanks!

PR #26, opened 2016-06-17
Elastic basic

montoyjh:
## Summary

* Basic elastic workflow that takes normal/shear deformations as inputs, meaning the user can set how many should be done
* Minor changes to VaspToDBTask that allow VASP documents to be connected to those from previous tasks in a workflow
* Store transformation_params in TransmuterFW

## TODO (if any)

* Tested Si, further tests will be done
* Is it possible to do change k-point parameters, like grid-density, for an entire workflow?  Should this be a power-up?

Also, as you may note from the commit log, @shyamd did most of the initial work on this, just wanted to acknowledge and thank him here.


computron:
Hi Joey,

Thanks very much for contributing this. A few notes before merging:

* Please change the code that looks like ``if "prev_task_id" in fw_spec: task_doc["prev_task_id"] = fw_spec["prev_task_id"]``. This is just adding some random new field from the fw_spec to the task doc, and if everyone did this, we would end up with lots of little things from the fw_spec automatically being put in the task doc. Instead, try to find a more general solution. For example, maybe introduce a new parameter called ``additional_fields_fw_spec`` where a user can give a list of keys in the fw_spec (like prev_task_id) and those will be added to ``additional_fields``. Then, the next time someone wants to add some other field from the fw_spec to the task doc, there will be an existing system to do so - they just state which fields from the fw_spec are important by using ``additional_fields_fw_spec``. I am open to other solutions so long as they can be of wider use than adding a single key to the task doc.

* ``not 'prev_calc_dir' in self`` can be written ``'prev_calc_dir' not in self``

As for your question about k-point density, usually the Fireworks have a way to control the k-point density used in that Firework (either via a parameter, or through the choice of inputset used). So the workflow (and particularly the automatic workflow) needs to set the appropriate parameters for each firework in the workflow. If you have any ideas for better solutions I am open to them.

montoyjh:
How about if the change is made such that fw_spec has a unique key that may be passed to the task_doc (I've used "task_doc_data") which can then be modified during the workflow, then there's still a single, general way to pass info from the fw_spec to the task doc.  I've added commits with the code I have in mind.

computron:
Btw I am going to change many of these commits. e.g. again the VasptoDbTaskDrone is passing arbitrary keys (task_doc_data.prev_task_ids) as part of its operation. I am removing this along with a few other things.

Note that the point of MatMethods is to actually separate the operation of tasks. If you need to pass certain keys in order to make your workflow work, add a FireTask for that (like I did with PassCalcLocs). The task for entering things into a database should do *only* that - enter things into a database. It should not be responsible for passing special keys needed to make certain workflows work.

PR #29, opened 2016-07-03
Just a test/practice for adding mpid to materials collection.

albalu:
I chose materials_ehull as it already calls Rester. Also, not sure if upsert is appropriate and whether ['N/A'] is necessary. I think it is more relevant to have this in tasks_materials.py but wasn't sure if I should call MPRester in that class so that I could use find_structure.

### DIFF
--- a/matmethods/vasp/builders/materials_ehull.py
+++ b/matmethods/vasp/builders/materials_ehull.py
@@ -40,10 +40,15 @@ class MaterialsEhullBuilder:
                 for x in ["is_hubbard", "hubbards", "potcar_spec"]:
                     params[x] = m["calc_settings"][x]
 
-                composition = Structure.from_dict(m["structure"]).composition
+                structure = Structure.from_dict(m["structure"])
+                try:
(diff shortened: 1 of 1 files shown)


computron:
Are you sure this is the correct merge request? This merge request seems to not do anything functionally different than the existing code.

If you just want to practice, you can do it in your fork (and ask for feedback there), but you should not submit a merge request until you feel the code is "ready to go".

albalu:
Thank you very much Anubhav for your comments. I tried to correct things as you said. I tested this commit and it works fine. BTW is there a case where find_structure() would find more than one id? I can't think of any as each "mp-..." material has a very specific composition, structure, volume, etc. If always len(mpids)==1, should I just add mpid as mpids[0] ?

computron:
Ok I am going to pull.

Note that you should leave it as an array. As long as the builders have the same definition of "material" as MP, then it should be 1:1. But if we start changing the criteria (e.g., we adopt looser standards for a single material than MP), then it may no longer be 1:1 and multiple MP ids might map to the same material as we've defined it.

PR #30, opened 2016-07-05
elastic workflow update

montoyjh:
## Elastic workflow update

This update adds a (minimal) analysis task back into the standard workflow, and provides some additional bookkeeping infrastructure.

* modify_task_docs vasp powerup allows users to modify all task documents generated by a workflow via the "additional_fields" parameter
* elastic workflow does an initial elastic tensor fitting with the 24 deformed tasks as parents.  Allows fizzled parents, so if one of the deformed structure tasks fizzles and is rerun, should also rerun.  Also will fail if fewer than 6 independent deformations succeed.
* Adds k-point "reciprocal density" parameter to workflow constructor.  

## TODO (if any)

* The k-point grid specification is done a bit hackishly at the moment, due to issues I had getting the grid_density to persist through serialization.
* Some preliminary tests have been unsuccessful at reproducing data generated via MPWorks, so use with caution.  Further testing is underway.


computron:
Hi Joey

For now I am just going to pull. I might do some tests of the elastic workflow and suggest modifications at a later point.

Thanks!

PR #31, opened 2016-07-06
Minor revision to make defuse_children optional

aykol:
- Helps control whether VaspToDbTask will check to defuse children fws based on whether the Vasprun "state" is "successful" or not. Defaults to True (i.e. existing behavior) and should not affect other existing workflows.
- In MDWF converged_ionic will always be False in Vasprun, and therefore the "state" will be "unsuccessful". Current revision  helps circumvent defused fws caused by this.


computron:
Hi Murat,

Can you just rename "check_defuse_children" to "defuse_unsuccessful"? I think the latter might be clearer. Other than that I think it is good to go, so I will pull as soon as I get the change.

aykol:
Hi Anubhav,
Just made that revision. Thanks!

PR #38, opened 2016-08-29
Dev elastic

montoyjh:
## Summary

Minor changes to elastic workflow to allow for additional, non-independent deformations (for higher-order tensor fitting) and a flag to allow for including the analysis task or not.


computron:
thanks!

PR #39, opened 2016-09-09
add_trackers powerup

albalu:
## Summary

I tried to make this more general @computron , could you please take a look at it? I'm running a test on fw_alireza to make sure std_err.txt is tracked but it hasn't started running yet.

## TODO (if any)

have a better control on nlines (maybe warning instead of exception?)


computron:
I think it is mainly good apart from two things:

1. Major Python gotcha - http://docs.python-guide.org/en/latest/writing/gotchas/
2. The assert statement is not necessary, let's not police people too much

albalu:
Ok I think I fixed it. Thanks for the link!

PR #40, opened 2016-09-10
Gibbs free energy workflow

matk86:
## Summary

combines dft data and quasi harmonic approximation to compute the gibbs free energy.

*  uses phonopy's quasi harmonic approx implementation.
*  unlike the “GIBBS” quasiharmonic Debyemodel, this one uses the computed phonon density to evaluate the free energy and entropy.
* updated the drone to assimilate dynamical matrix as well as the normal modes if present.
* The final analysis step can be easily extended to compute grueisen coeff as well as other paramters 
supported by phonopy.


computron:
Thanks -

Before I merge, can you avoid putting things like this:
```
try:
    from phonopy import Phonopy
    from phonopy.structure.atoms import Atoms as PhonopyAtoms
    from phonopy import PhonopyQHA
except ImportError:
    print("Install phonopy. Exiting.")
    sys.exit()
```

in the root of the module? Simply reading the module will cause the program to exit. For example, if FireWorks tries to read the module to search for a FireTask, it will encounter this code and will exit.

matk86:
sorry, forgot to move that to the function... fixed it

computron:
Thanks!

PR #41, opened 2016-09-10
Dev elastic

montoyjh:
Updated the elastic workflow to allow for custom k-points.

### DIFF
--- a/matmethods/vasp/firetasks/write_inputs.py
+++ b/matmethods/vasp/firetasks/write_inputs.py
@@ -69,7 +69,6 @@ class WriteVaspFromIOSet(FireTaskBase):
                                  self["vasp_input_set"])
             vis = vis_cls(self["structure"],
                           **self.get("vasp_input_params", {}))
-
         vis.write_input(".")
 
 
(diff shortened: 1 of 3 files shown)


matk86:
Hi Joey,
just saw your pull request to pymatgen. If you could add the volume dependent  debye temperature formulation as mentioned in [http://doi.org/10.1016/j.comphy.2003.12.001](url) we could easily extend your elasticity workflow(add one more task) to compute the gibbs free energy using the quasi harmonic debye model mentioned in the aforementioned paper. This would much more cheaper than computing the free energy from the full phonon spectrum.

montoyjh:
Sure, is this just the expression in equations 6-7 of that paper?

-Joey

matk86:
yes, but i am not sure how reliable those equations are ..the author cites his thesis as the reference but i couldnt find it..

montoyjh:
I've added it and a unit test (I get a reasonable number for the test
structure, Tin (around 189 K), but I'm really not confident that what I did
was correct, so we should test it more thoroughly before using it in
production.

matk86:
Sure, thanks

PR #45, opened 2016-09-25
attempt to add an soc powerup

albalu:
## Summary

attempt to add a vasp powerup to transform wfs/fws to spin-orbit coupling (SOC) calculations. At this point I've made some assumptions about nbands (passed as an input) and magmom to be 0.6 as it has been for all the calculations that I have done. It is a quick hack at the moment but I'm planning to make it more versatile in the future
- use vasp_ncl for running
- modify INCAR to have necessary SOC calculation flags
## TODO (if any)

*MAGMOM and NBANDS need to be assigned automatically, at least by default
- Another idea is to have a script that add static and nscf line and boltztrap with SOC inputs as children of a firework that has already been ran

@computron @matk86 , could you please comment. Also, I'm still new to the structures of the workflows, fireworks, etc so excuse me if this is not the best way of writing this powerup.

### DIFF
--- a/matmethods/feff/workflows/eels.py
+++ b/matmethods/feff/workflows/eels.py
@@ -21,7 +21,7 @@ logger = get_logger(__name__)
 def get_wf_eels(absorbing_atom, structure=None, spectrum_type="ELNES", edge="K", radius=10.,
                 beam_energy=100, beam_direction=None, collection_angle=1, convergence_angle=1,
                 user_eels_settings=None, user_tag_settings=None, feff_cmd="feff", db_file=None,
-                metadata=None, use_primitive=False):
+                metadata=None, use_primitive=False, feff_input_set=None):
     """
     Returns FEFF ELNES/EXELFS spectroscopy workflow.
 
@@ -30,8 +30,8 @@ def get_wf_eels(absorbing_atom, structure=None, spectrum_type="ELNES", edge="K",
         structure (Structure): input structure. If None and mp_id is provided, the corresponding
             structure will be fetched from the Materials Project db.
         spectrum_type (str): ELNES or EXELFS
-        edge (str): absorption edge
-        radius (float): cluster radius in angstroms
+        edge (str): absorption edge. K, L1, L2, L3
+        radius (float): cluster radius in angstroms. Ignored for reciprocal space calculations
         beam_energy (float): the incident beam energy in keV
         beam_direction (list): incident beam direction. Default is none ==> the spectrum will be
             averaged over all directions.
@@ -45,6 +45,7 @@ def get_wf_eels(absorbing_atom, structure=None, spectrum_type="ELNES", edge="K",
         use_primitive (bool): convert the structure to primitive form. This helps to
             reduce the number of fireworks in the workflow if the absorbing atoms is
             specified by its atomic symbol.
+        feff_input_set (FeffDictSet)
 
     Returns:
         Workflow
@@ -57,6 +58,7 @@ def get_wf_eels(absorbing_atom, structure=None, spectrum_type="ELNES", edge="K",
 
     override_default_feff_params = {"user_tag_settings": user_tag_settings}
 
+    # add firework for each absorbing atom site index
     fws = []
     for ab_idx in ab_atom_indices:
         fw_metadata = dict(metadata) if metadata else {}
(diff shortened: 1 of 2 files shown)


matk86:
Have you tested it? Also, regarding the documentation, please try to be consistent with the rest of the code(google style docstrings).

computron:
You need both elements and nelements for querying

The nelements is needed for querying ranges of element, e.g. nelements >4

https://docs.mongodb.com/manual/reference/operator/query/size/

albalu:
Ok you are right. I added back nelements. I currently have formation_energy_per_atom under thermo do you think it's ok there or it's more relevant under stability?

Also, Ge SOC test run is still waiting.

matk86:
it seems like your powerup replaces all vasp_cmd with vasp_ncl. It is inefficient(well, depends on system size) because vasp_ncl uses the full complex to complex fft(double the memory and slower) for charge densities and potentials. Optimization and static steps dont need vasp_ncl..

albalu:
Sure I can exclude the optimization step, but I think static needs vasp_ncl. The standard SOC is to generate CHGCAR with SOC and do an SOC nscf on that (for one thing, we need more NBAND for SOC calculations and it needs to be the same for static and nscf)

matk86:
static dont have to be non-collinear + if the non-collinear is started from chgcar with magnetization the magmom provided in the incar is ignored.. i did the tests a long time ago(i remember having some issues with chgcar), please verify this with the vasp wiki pages on LSORBIT and LNONCOLLINEAR...

albalu:
The nscf line and uniform tests on Ge with SOC has finished  but none of the INCAR flags are passed as if they weren't there. When I do lpad get_fws -i 5147 -d all, I see the dictionary of the INCARE flags I wanted to add under spec._tasks (see below) but it didn't actually do anything. Any idea why this happened?

            {
                "MAGMOM": "0 0 0.6 0 0 0.6 0 0 0.6 0 0 0.6 0 0 0.6 0 0 0.6 0 0 0.6 0 0 0.6 ", 
                "NBANDS": 128, 
                "LSORBIT": "T", 
                "LMAXMIX": 4, 
                "_fw_name": "{{matmethods.vasp.firetasks.write_inputs.ModifyIncar}}"
            },

computron:
@matk86 @albalu 

Have you guys resolved the procedure for doing SOC? I am no expert here but as far as I can tell:

1. Doing relaxation with SOC is technically possible, but probably unlikely to make a large difference while adding significant computational cost. One paper that does this is:
http://link.springer.com/article/10.1007/s11664-014-3276-1
They don't find very significant differences when running relaxation with SOC
2. Whether to do SCF in conjunction with SOC seems a bit tricky based on looking up some VASP literature. As far as I can tell:
- doing SOC with SCF on is going to be expensive, but there is more chance of finding the correct easy axis for magnetization since your saxis can rotate.
- doing SOC with SCF off (what is currently implemented in MatMethods) is OK but it is much better if you know in advance your magnetic easy axis *or* if you expect this to not matter much, e.g. if the compound is cubic or something. If there is a large degree of magnetic anisotropy I don't think you can randomly just choose the default saxis and run with SCF off since you will be unsure if you got the easy axis for saxis. You can try multiple axes to map out the degree of anisotropy:
https://wordpress.nmsu.edu/tmanz/files/2014/05/GabaldonLimas_Manz_2016_RSC_Advances.pdf

@matk86 , I didn't do any tests but I am not sure of this statement:
" if the non-collinear is started from chgcar with magnetization the magmom provided in the incar is ignored.."  That said, if it is true then maybe you just want to try different saxis and not mess with MAGMOM.

@albalu regarding your last message, I am pretty sure this was some way you were setting up FireWorks that was incorrect and that we fixed this awhile ago.

Best
Anubhav

computron:
Btw it seems for most materials the magnetic anisotropy is going to be small
http://cms.mpi.univie.ac.at/wiki/index.php/Determining_the_Magnetic_Anisotropy

So maybe just doing a single NSCF with an arbitrary (0 0 1) saxis is "good enough" most of the time to model spin-orbit behavior, unless you are really targeting magnetically anisotropic materials or want to be super careful

matk86:
@computron that's what I observed when I ran the tests long time ago and sure, I agree with your suggestions.

I have no problem with the changes(I'm not doing any mag vasp calculations  anyway) . I was waiting for @albalu to confirm  that his version runs without any error

PR #46, opened 2016-10-06
Feff tscc

czhengsci:
## Summary

1. Create TSCC feff submit firetask in firetasks/run_calc_tscc.py
2. Add a new firework in to fireworks/firework_tscc.py
3. Add a new EXAF test workflow into /workflows/exafs.py

## TODO (if any)

Detailed document
And test

### DIFF
--- a/matmethods/feff/firetasks/run_calc_tscc.py
+++ b/matmethods/feff/firetasks/run_calc_tscc.py
@@ -1,0 +1,139 @@
+# coding: utf-8
+
+from __future__ import division, print_function, unicode_literals, absolute_import
+
+"""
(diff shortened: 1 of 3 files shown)


matk86:
you dont need run_calc_tscc.py, its basiaclly reusing stuff from fireworks, instead you could wrap all the feff mpirun calls in a bash wrapper script and add that your path(or set the path in your fireworks config file). I have emailed you the bash script that i wrote for nersc, just replace srun with mpirun...once you have that you dont need custom exafs workflow for tscc, you could just use the one i wrote, just replace the feff_cmd option...

matk86:
closing this pr

PR #47, opened 2016-10-07
Feff

czhengsci:
## Summary

*add db module for eels database management
*add FEFFWorkflowManager in fireworks/firework_tscc for job submission on tscc side

## TODO (if any)

*Chen need to complete document of different modules
*Chen need to expand FEFFWorkflowManager to other spectrum calculations

* workflow support EXAS job submission now


matk86:
@czhengsci we dont need run_calc_tscc.py, please see my comment on your previous pull request(#46).
I was able to run the workflow as it is on tscc with no glitches..

PR #48, opened 2016-11-04
Fix the Debye model GibbsFreeEnergyTask

bocklund:
## Summary

* `vasp.firetasks.parse_outputs.GibbsFreeEnergyTask`
    - Changed the force constants to not be read when using `qha_type="debye_model"`. The Debye model version of the Gibbs free energy task would fail because the force constants were being read from the database documents, but they are not calculated when the Debye model is used and are not in the database.

* `tools.analysis.get_debye_model_gibbs`
    - Changed number of elements in linspace to be `(t_max-t_min)/t_step +1` to have enough elements for a closed loop. I decided to keep `np.linspace` over `np.arange` to better support floating points.
    - Return values for the function were changed to simply be lists rather than NumPy arrays. Arrays are not needed and are not properly converted to lists by `json.dumps`.


computron:
Thanks!

PR #49, opened 2016-11-07
Minor bug fix

montoyjh:
## Summary

 Minor fix to enable Static FWs to use user-specified VIS, reversed the relevant `or` expression


computron:
Thanks!

PR #51, opened 2016-11-11
Fix phonon calculation analysis

bocklund:
## Summary

* Previous changes moved the `force_constants` processing into another loop for phonon calculations to separate the processing from the Debye model. This broke the phonon post-processing because the `force_constants` were being appended to a list of outputs instead of applied to the ones that already existed. 
* Moving the `force_constants` processing back into the loop fixed this. 
* The solution is perhaps not the most elegant, but matches the style of existing code and has little duplicated code.

## TODO (if any)

* Write tests with mocks for the analysis routines to make sure changing one thing does not break another. This is especially important for analysis tasks that have multiple modes such as phonon calculations and Debye calculations
* Think about how to best check for the `qha_type`. I think we can do better than just checking if the `qha_type` is `"debye_model"` or not.

### DIFF
--- a/matmethods/vasp/firetasks/parse_outputs.py
+++ b/matmethods/vasp/firetasks/parse_outputs.py
@@ -388,9 +388,13 @@ class GibbsFreeEnergyTask(FireTaskBase):
         for d in docs:
             s = Structure.from_dict(d["calcs_reversed"][-1]["output"]['structure'])
             energies.append(d["calcs_reversed"][-1]["output"]['energy'])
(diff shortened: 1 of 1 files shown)


matk86:
are you saying that the previous changes that you made broke it? Please make sure that you test the changes before making the pull request.

bocklund:
That's correct. I tested the Debye version only, which is why I did not catch the mistake. I have now tested both and they are working correctly.

matk86:
ok thanks.. could you also provide some unittests to prevent further regressions? thanks

PR #53, opened 2016-11-20
updating modify_to_soc and tag_fws powerups both tested and working 

albalu:
## Summary

updating modify_to_soc and tag_fws powerups both tested and working

## TODO (if any)

This powerup is for when a workflow is just defined and no previous calculations are available. For former case, current MatMethods SOC Firework should be modified to do both a static (with vasp_cmd) and nscf SOC (with vasp_ncl) so NBANDS MAGMOM are read from previous calculations

### DIFF
--- a/matmethods/vasp/powerups.py
+++ b/matmethods/vasp/powerups.py
@@ -12,6 +12,8 @@ from matmethods.vasp.firetasks.run_calc import RunVaspCustodian, RunVaspDirect, 
 from matmethods.vasp.firetasks.write_inputs import ModifyIncar
 from matmethods.vasp.config import ADD_NAMEFILE, SCRATCH_DIR, ADD_MODIFY_INCAR
 
+from pymatgen import Structure
+
 __author__ = 'Anubhav Jain, Kiran Mathew'
 __email__ = 'ajain@lbl.gov, kmathew@lbl.gov'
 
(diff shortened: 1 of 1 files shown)


computron:
@matk86 can you review and merge? You worked on the original SOC

@albalu - note that 0.6 magmom initialization is typically for non-magnetic elements only. For magnetic elements we typically initialize with magmom=5. The pymatgen input sets probably have this listed somewhere.

matk86:
1. this will convert all non structure optimization and non static fireworks to soc irrespective of what the fw constraint is?

2. why is the task_name_constraint set to "RunBoltztrap" ?

3. shouldnt saxis be a parameter too?

albalu:
@matk86 :
1. I made a correction, now all RunVasp that are not structure will be with SOC and only nscf ones with be run with vasp_ncl
2. That one is just to add soc at the end of RunBoltztrap tasks to differentiate with regular Boltztrap as they may be different and they are run with the soc=True argument in RunBoltztrap
3. Yes, somehow saxis in INCAR would be printed as 0 only even though I had set it as (0, 0, 1); for now, I let VASP use the default value which is that but for later I will test and implement this correctly. I added a TODO to do this later.

albalu:
sorry I might have accidentally pushed in the main branch before but it's the same changes as here so the merge shouldn't have any conflicts.

matk86:
ok. the mpsocset in pymatgen.io.vasp.sets handles saxis properly

albalu:
Thank you. Actually I only made nscf runs SOC (with LSORBIT = T, etc); this is much faster as it doesn't require static run to be with vasp_ncl as you said before. I made a new commit. Do I need to open a new pull request?

matk86:
sure, or you could make a pr after doing the saxis fix too.. up to you..

albalu:
Ok I just did; thanks for your patience. I will do the saxis later, I have to go to sleep now.

PR #55, opened 2016-12-02
including the field "scissor" in boltztrap db

albalu:
## Summary

reading the value of "scissor" that is applied to the band structure (0.0 is the default). This is needed to differentiate between boltztrap runs with a scissored band structure and the ones with a normal band structure of the same materials. "scissor" is not available through BoltztrapAnalyzer so I added a function to read it from files.

### DIFF
--- a/matmethods/vasp/firetasks/parse_outputs.py
+++ b/matmethods/vasp/firetasks/parse_outputs.py
@@ -142,6 +142,15 @@ class BoltztrapToDBTask(FireTaskBase):
 
     optional_params = ["db_file", "hall_doping"]
 
+    def get_intrans(self, btrap_dir):
+        intrans = {}
+        with open(os.path.join(btrap_dir, "boltztrap.intrans"), 'r') as f:
+            for line in f:
+                if "iskip" in line:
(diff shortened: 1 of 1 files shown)


albalu:
@computron, could you please comment? I need this to run a new set of boltztrap runs with a fixed gap (maybe 1 eV).

Also, note that another way of differentiating between scissored and normal boltztrap runs is to use "tags" but they are not passed to the boltztrap collection.

computron:
Hi Alireza

The code to parse scissor should be in pymatgen rather than MatMethods. MatMethods is not responsible for file parsing and I/O. If we start mixing this functionality into MatMethods, things will quickly start being a mess. You should instead:

1. Fork pymatgen
2. Make the changes to parse the scissor value in the BoltztrapAnalyzer in pymatgen. Make sure the as_dict() function also outputs the scissor value
3. Merge request the fork into main pymatgen and ask for a version update soon.
4. Update the requirements.txt and setup.py in MatMethods to include the latest pymatgen version that includes the change.

This way, you require zero changes to MatMethods itself.

Also note that internally, you don't need to wait for the version update. Just update your pymatgen compilation to the latest master after it is merged.

computron:
@albalu I am closing this PR for now. Feel free to reopen if needed

PR #56, opened 2016-12-08
Unit test elastic

montoyjh:
## Summary

Working unit test for the elastic workflow

## TODO (if any)

Abstraction into unittests with and without running workflow/vasp might eventually be implemented.


computron:
Thanks!

PR #57, opened 2016-12-08
MatMethods test suite

montoyjh:
## Summary

Adds functionality to mmwf to add a test_suite.

* Test suite can be added to LaunchPad by issuing `mmwf test`.  LaunchPad can be reset by using `mmwf test --reset`.
* Few minor edits involving the elasticity test from my last PR.

## TODO (if any)

* Test set could be more well-thought-out (i. e. better coverage, more appropriate structures for certain workflows, some workflows might not need testing, etc).
* Only tests vasp wflows as of right now.  Maybe feff/lammps etc. should be added?
* Reset should maybe clean vasp db of test calcs as well.
* verification of test set.


computron:
Let me know when you have a basic verification of the tests

montoyjh:
Should be good to go.

computron:
Thanks! Can you make the tests not depend on MPRester? just store the Structures somewhere. Otherwise MM unit tests will fail when MP is down, and also other problems in MP will fail MM unit tests.

shyuep:
Why don't you use PymatgenTest.get_structure? Then you do not need MPRester.

computron:
@shyuep That sounds good to me, so long as @montoyjh can add a few extra structures if needed into PymatgenTest

montoyjh:
Agreed.  I'll change it back, I think I might need to add one structure to
the test structures.

montoyjh:
Should be good.

PR #58, opened 2016-12-09
Adding a full HSE band structure workflow

albalu:
## Summary

The implementation of MPHSEBSSet.from_prev_calc only allowed adding the CBM and VBM, this was modified independently to also add all the high-symmetry points to calculate a correct HSE band structure.

On the Matmethods side I modified the current infrastructure just a little bit to include "mode" as an input for HSEBSFW controlled via gap_only (mode=="uniform") or not gap_only (mode=="line")

## TODO (if any)

Have a set of tasks to split a full HSE band structure to several jobs so one big HSE jobs doesn't crash due to memory.


computron:
Thanks! 

A few notes before merge

1. You changed the name of "hse gap" to "hse ". I think this is a typo but please fix it! task_labels matter in MatMethods and messing them up can have big consequences

2. Other methods in pymatgen use lowercase-only (not first-letter capital) for things like "uniform" and "line". you should do the same, don't invent a new convention (see the section of the code where you call capitalize()

3. Not sure why you are changing "uniform" to "gap". Might as well have the user enter "gap"? It is confusing if it is one name for the user ("uniform") and another name within the code ("gap"). Keep these things consistent, otherwise debugging becomes more difficult

computron:
@albalu 

Actually for point #1 it looks like pymatgen sometimes uses capital-case for "Uniform" (e.g. MPHSEBSSet), other times lower case for "uniform" (e.g. MPNonSCFSet). Please coordinate with @shyuep to see which one he wants to use throughout pymatgen. Also check "Line" vs "line"

albalu:
Hi @computron , 

for #1. actually the name would be hse gap if mode=="uniform" (like before) and hse line if mode=="line" so it is never just "hse"

for 2. in Matmethods we have been using "uniform", if I change that we will have to change **every** occurrence of "line" and "uniform" to "Line" and "Uniform" throughout matmethods which is fine by me but just giving you a headsup. Please decide on a convention (i.e., Line vs. line, etc and I will stick to it)

3. Are you referring to the name of the workflow or the function input? I can just change everything to line and uniform and add in the documentation that uniform includes the VBM and CBM k-points anyway.

computron:
@albalu 

1. Ok I see what you are doing now, but it's kind of confusing that the "name" parameter is actually now a *prefix* and the code automatically appends "gap" or "line" to the end (regardless of whether the user wants this to be appended or not! And btw if the user forgets the trailing space, then they might end up with something like "hse aj_customline"). I would suggest setting the name = None by default in the method signature. Then in the code, if the name is not already set to whatever the user wants it to be, you can set it to "hse gap" or "hse line" depending on the setting

2. Please coordinate with @shyuep on the pymatgen guideline. For MatMethods we will copy it throughout to be the same as pymatgen. It is OK if this involves many changes. Let's just stick with one thing. Note that I tend to just prefer lowercase. For one thing, it reduces ambiguity when there are (potentially) titles like "uniform and dense". Otherwise, is it "Uniform and Dense", "Uniform And Dense", or "Uniform and dense"? Another thing is that it saves me having to get the capitalization correct when typing. But I will take consistency over my preference.

3. The name of the "mode" is "uniform" but the name of the task is "hse gap". It is better if mode is "gap" since you are not actually doing a uniform mode calculation (in fact, you are breaking uniform mesh by adding two potentially off-symmetry k-points). Similar for pymatgen. This will leave room for a truly HSE uniform task that computes a dense uniform k-point mesh.

albalu:
1. Ok that's better. I put None as default and define the default names based on the mode. I admit I didn't put much thought into this part.

2. @shyuep, can I just make "Uniform" and "Line" lower case also in MPHSEBSSet and make a pull request?

3. Ok how about three modes: uniform, gap, line ? Please reply if you object.

shyuep:
I would suggest simply making it such that comparisons are not case sensitive.

computron:
1. ok i will wait for the change
2. given @shyuep 's comment, let's use lowercase within our code with the stipulation that comparisons are not case sensitive (i.e., if the user were to completely mangle the capitalization, it would still be OK). Note that for things like task_label in MatMethods we need to actually pick a capitalization strategy, and since we have been using lowercase throughout, we can stick to that.
3. having 3 modes is fine with me, as long as they all make sense. i.e. I am not sure how Uniform mode works with HSE given the whole issue with doing NSCF HSE

shyuep:
Done in https://github.com/materialsproject/pymatgen/commit/26f99797b942a98246a831b30baba9de2e4302d0

albalu:
2. Thanks @shyuep 

3. the name uniform itself is a little confusing to me as it may imply that it is an nscf step which is NOT. All these modes are self consistent hse. I tested and even if an nscf step is done after this static step with the same k-points etc, the same band structure is resulted. Perhaps the default modes: static, static gap, static line may be more clear? and the Wflow names would be hse static, hse static gap, hse static line. If you disagree with this I would just implement the uniform, line, gap mode now.

computron:
For 3) just do line and gap - those are the two that require playing special tricks with the k-point mesh. Forget the uniform. For static one can probably just use the MPStaticSet and override the HSE params. We can worry about that one later

Anubhav

albalu:
Ok I have the gap and line options now. mode=="unifrom" can still be passed anyway as the MPHSEBSSet has such an option that does an scf calculation only on the uniform k-mesh.

computron:
Thanks!

PR #59, opened 2016-12-16
Control of soc and scissor options BoltzTraP + VaspToDbTask

albalu:
## Summary

* BoltztrapRunner on a band structure that is run with spin-orbit coupling requires the argument soc to be set to True, so I added this option explicitly to BoltztrapFW , etc

* It turns out scissor is actually the final band gap, so I made this clear in the documentation and added this to BoltztrapFW as well for more control.

### DIFF
--- a/matmethods/vasp/firetasks/parse_outputs.py
+++ b/matmethods/vasp/firetasks/parse_outputs.py
@@ -58,10 +58,8 @@ class VaspToDbTask(FireTaskBase):
         defuse_unsuccessful (bool): Defuses children fireworks if VASP run state
             is not "successful"; i.e. both electronic and ionic convergence are reached.
             Defaults to True.
-        check_all_files (bool): whether to check for existence of all the files that VASP
-            normally creat upon running. Defaults to True.
     """
-    optional_params = ["calc_dir", "calc_loc", "parse_dos", "bandstructure_mode", "check_all_files",
+    optional_params = ["calc_dir", "calc_loc", "parse_dos", "bandstructure_mode",
                        "additional_fields", "db_file", "fw_spec_fields", "defuse_unsuccessful"]
 
     def run_task(self, fw_spec):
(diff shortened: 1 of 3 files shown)


albalu:
I also modified VaspToDbTask to, by default, check for all VASP files to exist before parsing to make sure VASP actually started running. This can be disabled via check_all_files=False in case the user is confident that they only want to use vasprun.xml, OUTCAR, etc.

If the implementation is ok I can add a unittest for this.

computron:
Hi Alireza

1. the implementation check_all_files (default=True) is just asking for trouble. People like to turn off WAVECAR printout, for example (or delete WAVECARs to save disk space - similar with CHG/CHGCAR), and this will cause failure. Please only check for VASP files that are always present. There is no need to enforce things like WAVECAR/CHG/CHGCAR to be there.

2. Also, it is better if you implement the checks as a Validator for custodian rather than in VaspToDbTask. That is the place where someone would expect validation of the run.

3. Note that it is possible to parse runs that are incomplete and put them in the database (the state is usually "failed" rather than "successful".) So again when validating files, make sure you are only validating things that are really going to be there if an OUTCAR is there (e.g., stuff created when VASP first starts running).

albalu:
1. I can take out WAVECAR/CHG/CHGCAR from the checklist (as it is more likely for the user to remove these as opposed to say DOSCAR) but they are also automatically created upon running of VASP though they may be empty. 

2. It makes sense. Based on this, I would just revert the changes

3. All those files are created immediately after VASP starts running.

Anyway, based on 2, it seems like we don't need this 2nd check in VaspToDbTask as we discussed before as it should be implemented in Custodian after vasp starts running.

computron:
@albalu you need to resolve the conflicts, probably due to the rename of matmethods->atomate

computron:
@albalu note it might be simpler to just re-fork (now atomate), re-commit the changes manually into the atomate fork, and submit a new PR from that fork

albalu:
Yes that's what I was thinking.

PR #61, opened 2016-12-16
previous pull request to add soc and scissor to Btrap FW

albalu:
## Summary

soc and scissor can be tuned directly in BoltztrapFW for better control.


computron:
Thanks!

PR #64, opened 2016-12-21
add the ability to choose ObjectId in insert_gridfs

albalu:
## Summary

I may want to rebuild/update a collection (e.g. boltztrap + boltztrap DOS) but keep the old object ids and in a sense just update some of the documents w/o changing the ids, this small change to the code let me do that without changing its default functionality,

### DIFF
--- a/atomate/vasp/database.py
+++ b/atomate/vasp/database.py
@@ -64,7 +64,7 @@ class MMVaspDb(MMDb):
                                           ("completed_at", DESCENDING)],
                                          background=background)
 
-    def insert_gridfs(self, d, collection="fs", compress=True, _id=ObjectId()):
+    def insert_gridfs(self, d, collection="fs", compress=True, id=None):
         """
         Insert the given document into GridFS.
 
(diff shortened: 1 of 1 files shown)


computron:
Can you make the following changes:

1. Rename "_id" to "oid". Generally the leading underscore indicates things that a user should not change; thus, having a leading underscore variable be a parameter to a function is not really correct naming.
2. in the function signature, default oid to "None" (not ObjectId())
3. inside the function, you can add something like:
oid = oid or ObjectId()

This will set the default value to ObjectId() if it was originally None.

You might ask why not just set the default var to ObjectId() in the function header rather than the whole "None" business. The reason is because Python creates very terrible bugs if you put mutable objects inside function headers:

http://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

That said, ObjectId() should be non-mutable and you are most likely OK even if you do not do 2 and 3 on the list above. But with this particular Python "feature/issue" of mutable default args I tend to be very cautious and only put things that are very clearly immutable as default vars.

albalu:
@computron 
1. done. Thanks!
2,3. that's how I first wrote it but I wasn't sure it applies to sth like ObjectId() as well. I will keep this in mind.

computron:
Thanks!

Note that "id" is a built-in Python function and you should avoid overlap (Pycharm should warn you about this name by drawing a squiggly underneath the name) - oid is better because it does not conflict with any Python functions.

Also, for 2,3 I am >90% sure that the initial implementation was OK due to immutability of ObjectId. But again I am just paranoid about this.

PR #66, opened 2017-01-04
formation energy per atom in builder + more general tag_fws

albalu:
## Summary

1. I implemented formation energy per atom in the Ehull builder before so it's tested but I forgot to merge it before.

2. users now have more control over tag_fws to tag the entire workflow and not just VaspRun jobs

### DIFF
--- a/atomate/vasp/builders/materials_ehull.py
+++ b/atomate/vasp/builders/materials_ehull.py
@@ -52,6 +52,13 @@ class MaterialsEhullBuilder(AbstractBuilder):
                 self._materials.update_one({"material_id": m["material_id"]},
                                            {"$set": {"stability": self.mpr.get_stability([my_entry])[0]}})
 
+                for el, elx in my_entry.composition.items():
+                    entries = self.mpr.get_entries(el.symbol)
+                    min_e = min(entries, key=lambda x: x.energy_per_atom).energy_per_atom
+                    energy -= elx * min_e
+                self._materials.update_one({"material_id": m["material_id"]},
+                                           {"$set": {"thermo.formation_energy_per_atom": energy / structure.num_sites}})
+
                 mpids = self.mpr.find_structure(structure)
                 self._materials.update_one({"material_id": m["material_id"]}, {"$set": {"mpids": mpids}})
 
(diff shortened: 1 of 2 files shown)


albalu:
@computron, could you please take a look at this? they are small changes.

computron:
For (1), you need to use the Advanced Compatibility framework to correct the raw energies. Otherwise you will not get an accurate formation energy for many compounds, e.g., oxides.

For (2), the optional task_level_constraint is fine, but this is technically a backward-incompatible change since the old default ("RunVasp") is now None. Such changes must involve (i) a statement that all code usages of that function have been properly updated and (ii) a note that this is backwards incompatible, and users who called that function in downstream code likely need to change it. This change could break other people's existing code.

computron:
Note that

"a statement that all code usages of that function have been properly updated" involves actually finding all usages of that function (PyCharm can help you) and adding the old task_level_constraint to maintain the prior behavior.

computron:
Also by "Advanced Compatiblity" I meant "MaterialsProjectCompatibility" (a long time ago, we called it the former)

albalu:
Thanks Anubhav, 

1) it seems that the get_entries method already enforce MaterialsProjectCompatibility by default but I made sure that it is used by adding compatible_only=True when I call the method

2) I actually didn't think about backwards-compatibility, thank you. The easier way would have been to keep the task_name_constraint="RunVasp" as default but I think None makes more sense. So I updated all the usages of tag_fws in the local codes which are all private; where do I have to include "a note that this is backwards incompatible, and users who called that function in downstream code likely need to change it." ? Is this a warning inside that function or just a part of function doc?

computron:
Hi @albalu 

1. Ok! I didn't notice that about get_entries. In that case the formation energies should be correct. You might want to double check an oxide just to make sure

2. 
>> The easier way would have been to keep the task_name_constraint="RunVasp" as default but I think None makes more sense. 

yes I agree with both parts

>> So I updated all the usages of tag_fws in the local codes which are all private

ok good; although sometimes it's better to wait until the merge is in place before pushing the updates to the downstream codes

>> where do I have to include

In the pull request, you should make a note of it. We can discuss how to deal with the backwards incompatibility, whether it is to support the old way more fluidly or to make a decision to change things going forward. In this case, what I would normally do is just announce it on the Google group for the code plus on Slack when releasing the new version. Since atomate does not have a Google group and this is a relatively minor change I will skip the announcement this time.

PR #68, opened 2017-01-11
support terminate function for slurm and add relax choose to OptimizeFW

specter119:
## Summary

* add custodian job_type  choice to OptimizeFW
* add terminate_func for slurm to RunVaspCustodian

### DIFF
--- a/atomate/vasp/firetasks/run_calc.py
+++ b/atomate/vasp/firetasks/run_calc.py
@@ -51,8 +51,7 @@ class RunVaspDirect(FiretaskBase):
         vasp_cmd = env_chk(self["vasp_cmd"], fw_spec)
         logger.info("Running VASP using exe: {}".format(vasp_cmd))
         return_code = subprocess.call(vasp_cmd, shell=True)
-        logger.info(
-            "VASP finished running with returncode: {}".format(return_code))
+        logger.info("VASP finished running with returncode: {}".format(return_code))
 
 
 @explicit_serialize
--- a/atomate/vasp/firetasks/run_calc.py
+++ b/atomate/vasp/firetasks/run_calc.py
@@ -206,7 +206,8 @@ class RunVaspCustodian(FiretaskBase):
 
         c = Custodian(handlers, jobs, validators=validators, max_errors=max_errors,
                       scratch_dir=scratch_dir, gzipped_output=gzip_output,
-                      terminate_func=terminate_func, terminate_on_nonzero_returncode=terminate_on_nonzero_returncode)
+                      terminate_func=terminate_func,
+                      terminate_on_nonzero_returncode=terminate_on_nonzero_returncode)
 
         c.run()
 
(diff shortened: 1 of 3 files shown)


specter119:
Summary

add custodian job_type choice to OptimizeFW
add terminate_func for slurm to RunVaspCustodian

atomate/vasp/firetasks/run_calc.py
```diff
@@ -193,10 +245,12 @@ def run_task(self, fw_spec):
         doping = self.get("doping", None)
         soc = self.get("soc", False)
 
-        vasprun, outcar = get_vasprun_outcar(".", parse_dos=True, parse_eigen=True)
+        vasprun, outcar = get_vasprun_outcar(
```
bocklund: I think these sort of changes where you are splitting one line into 2 just to satisfy a line length limit does not make sense. The idea behind an 80 character limit is to help make code more readable and splitting these one line function calls to two lines undermines that, in my opinion.

computron:
Hi @specter119 

Thanks for submitting a PR, but there are a couple of issues here which is why I am closing the PR.

1 - For line length in the codebases I manage (fireworks, matminer, atomate), I would like to set 100 characters as a soft limit for line length and 120 of a more hard limit. The PEP guideline of 80 in my experience reduces code readability for lines in the 80-100 character range much more often than it helps. I agree with @bocklund that the specific changes here actually serve to reduce readability.

2 - The code for SLURM cancellation, etc., would have been fine in MPWorks, but the reason we created the new code atomate was to keep the workflow definitions as clean as possible. This means that in atomate, it is not allowed to mix commands pertaining to specific queuing systems, etc. within the workflow definitions. The workflows are all about materials science procedure and not about details of killing jobs properly. For the latter, you can either try to do this via custodian. Either write a handler and pass it in to the atomate workflows, or write a new custodian Job and pass that in to atomate. However, the Firetasks, Fireworks, and Workflow definitions should be as minimal as possible and directly related to the essentials of the calculation procedure.

Again, thanks for trying this, but we are doing some policy changes for atomate vs. MPWorks.

specter119:
@computron @bocklund Yes, I got your points, i will try to add it to custodian.

shyuep:
@computron @specter119 No, SLURM commands are not allowed in custodian. Custodian has no queue specific commands. Not even PBS ones. So a SLURM cancel command would be odd.

computron:
Ok, in that case I guess this must remain custom code for @specter119 that is not pushed to any of the main repos since I cannot think of a place for it. Queue specific things can go in FireWorks, e.g. in the queue adapters. I can see some benefit to having a terminate command supported by QueueAdapterBase and implemented specifically for the various systems (PBS, SLURM, etc.) although ideally this feature would somehow link into FireWorks capabilities.

specter119:
@computron @shyuep Thanks, I think the terminate function is indispensable to complete a workflow under SLURM. I just use my branch of atomate as a temporary solution. But I have no idea how to add it to FireWorks right now.

bocklund:
I think it could be relatively straightforward to implement a cancel function for the queues already defined in the `fireworks/user_objects/queue_adapters/common_adapter.py`, but I defer to @computron. Out of curiosity, what's the use case for automated cancellation?

PR #69, opened 2017-01-12
more control over materials builder + add_tags also through BoltztrapFW and BoltztrapToDBTask

albalu:
## Summary

1. user can specify some tasks with excluded_tasks argument (a query) so they will be excluded when "materials" collection is being built. Examples of excluded_tasks is {"tags": "test"} or {"tags": {"$all": ["ABCD3", "soc"]}}

In the first example, I wouldn't want the "test" tasks to be included in the final materials collection and in the second one, I don't want the energy values of spin orbit coupling calculations for a specific projects to be included when say energy above hull is calculated.

2. now add_tags also tags Boltztrap jobs so organizing them would be easier later. A unit test is added but I also tested this on a bunch of bandstructure_boltztrap Wflows and the tags show up in the boltztrap collection.

### DIFF
--- a/atomate/vasp/builders/tasks_materials.py
+++ b/atomate/vasp/builders/tasks_materials.py
@@ -26,7 +26,8 @@ module_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)))
 
 
 class TasksMaterialsBuilder(AbstractBuilder):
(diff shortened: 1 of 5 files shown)


computron:
Thanks!

PR #70, opened 2017-01-25
Adding the option to turn off initial structure optimization in elastic workflow.

jdagdelen:
Gives user the option to skip the initial optimization task in the elastic workflow. Preset to perform optimization first.


computron:
Thanks!

Two suggestions:

* optimize_first -> optimize_structure
* move all the vis_relax code inside the if statement where possible. i.e., you are only executing necessary code when optimize_structure=False

computron:
Thanks!

PR #72, opened 2017-01-26
Update common powerups for vasp

### LINKED ISSUE #71: No gamma_vasp in default workflows
None of the default workflows populate vasp_gamma_cmd in RunVaspCustodian so that doesn't get set. I can go ahead and modify the default workflows to set that or add in a powerup. Any preference?

shyamd:
Added gamma_vasp powerup using to fix #71 
Updated default configs to use env_check
Added in gamma_vasp to common powers


computron:
Thanks!

computron:
Thanks!

PR #74, opened 2017-02-13
feff updates: EXAFS scattering paths workflow

matk86:
* workflow to compute EXAFS scattering amplitudes for custom scattering paths.
* added unittests(not integration test) for the added firetasks, fireworks and workflow.


computron:
Thanks!

PR #77, opened 2017-02-22
Gives users the ability to do elastic tensor calculation in 6 rather than 24 calculations. 

jdagdelen:
## Summary

This change adds the options for users to easily create an elastic constant workflow that consists of 6 rather than 24 strain-stress calculations. This "quick calculation" workflow does not necessarily produce results consistent with the MP procedure and is not used for MP elastic tensor calculations. As before, if the user specifies the shear and/or norm deformations, those values are uses rather than the defaults. 

* Adds "quick_calculation" as possible key in config dict, default is False
* Default strains are 0.01 normal and 0.03 shear if quick_calculation = True
* Does not change behavior of wf_elastic_constant if user specifies strains

### DIFF
--- a/atomate/vasp/workflows/presets/core.py
+++ b/atomate/vasp/workflows/presets/core.py
@@ -197,6 +197,8 @@ def wf_elastic_constant(structure, c=None):
     if quick_calculation:
         norm_deformations = c.get("norm_deformations", [0.01])
         shear_deformations = c.get("shear_deformations", [0.03])
+
+    
     wf = get_wf_elastic_constant(structure, vasp_cmd=vasp_cmd,
                                  norm_deformations=norm_deformations,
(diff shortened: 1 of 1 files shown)


computron:
Hi @jdagdelen 

Shouldn't the number of calculations depend on the symmetry of the cell?

e.g., cubic has fewer deformations, triclinic has more. I think the issue with the "24 calcs" is not so much that there are 4 calculations per deformation mode (6 modes), but rather that in many symmetric systems there is perhaps only 1-3 relevant deformation modes. e.g., for cubic you would just need a single deformation mode.

jdagdelen:
The sense that I got from the discussion was that people wanted to fit through zero with one strain at each mode. I'm working on the symmetry-considering workflow as well. Maybe we'll just hold off until that's finished.

PR #78, opened 2017-02-23
change the name of tag_fws to set_fworker to avoid confusion

albalu:
change the name of tag_fws to set_spec_fworker to avoid confusion with the add_tags powerup. Also the variable "tag" is renamed to fw_queuename to better reflect the goal of this powerup. Please feel free to modify these names.

This change is backwards incompatible but I assume it has not been largely used. In my codes, at least, I can correct all the names later.


computron:
Can you just set the name to ``set_fworker`` instead of ``set_spec_fworker``?

albalu:
ok I did.

computron:
thanks! i'll probably also make a few minor edits

PR #79, opened 2017-02-24
Adsorption

montoyjh:
## Adsorption workflow tools

* Users can generate workflows for conducting calculations of adsorption energies on surfaces
* Two workflow-generating functions:
    - one which generates optimize fireworks with adsorbates on a pre-existing slab
    - second which begins with a structure optimization based on transformations
* Also adds a workflow to perform calculations of molecules in VASP for reference states, if desired
* Unittest/test_files which tests basic functionality and performs a simple validation of results

## TODO (if any)
* Adsorbate vibrations functionality
* Automated analysis based on user reference input


computron:
Thanks!

PR #80, opened 2017-02-27
modify FitEquationOfStateTask to insert the data to db rather than to .json

albalu:
## Summary

* db_file is a required parameter so we put the result in db rather than printing to a .json file
* in addition to EOS data, the structure optimization data (+ FW tags) is passed to the "eos" collection to make further queries easier.
* bulk modulus in GPa will be available in the root of each document within the "eos" collection as all EOSs have the parameter b0 (b0 is in eV A^-3)

## TODO:
* a test for wf_bulk_modulus that has this task at the end
* I kept the name of the class FitEquationOfStateTask so it doesn't have ToDb in its name; not sure if that's a problem.
* the name "eos" is selected for the collection as it has more information than just the bulk modulus; feel free to modify it

### DIFF
--- a/atomate/vasp/firetasks/parse_outputs.py
+++ b/atomate/vasp/firetasks/parse_outputs.py
@@ -451,33 +451,39 @@ class FitEquationOfStateTask(FiretaskBase):
 
         tag = self["tag"]
         db_file = env_chk(self.get("db_file"), fw_spec)
-        summary_dict = {"eos": self["eos"]}
+        eos_dict = {"eos": self["eos"]}
 
         mmdb = MMVaspDb.from_db_file(db_file, admin=True)
         # get the optimized structure
         d = mmdb.collection.find_one({"task_label": "{} structure optimization".format(tag)})
         structure = Structure.from_dict(d["calcs_reversed"][-1]["output"]['structure'])
-        summary_dict["structure"] = structure.as_dict()
 
         # get the data(energy, volume, force constant) from the deformation runs
         docs = mmdb.collection.find({"task_label": {"$regex": "{} bulk_modulus*".format(tag)},
                                      "formula_pretty": structure.composition.reduced_formula})
         energies = []
         volumes = []
-        for d in docs:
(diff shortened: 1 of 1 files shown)


computron:
@albalu Thanks but there should also be the JSON option. See VaspToDbTask where there is a db_file option that controls whether a file or a DB insertion is conducted.

albalu:
@computron but db_file is required here as the structure is looked up from tasks so having the JSON option is redundant. I can do one of the following:

1. don't change anything as db_file is required in FitEquationOfStateTask in its current form.
2. make db_file optional and whenever db_file is not available, instead of tasks collection, get the location of the first FW (structure optimization) and get its VaspDrone and pass that (+bulk modulus, etc) on to "eos"
3. do 2 either db_file is available or not.

which one should I do? I prefer 3 as the "tasks" collection may not be available for looking up the structure optimization task. Besides, it is safe to assume that the first FW is the structure optimization from the initial structure and not a deformed structure.

computron:
@albalu You could add an optional parameter to output to JSON (default false) which bypasses the db insertion and instead outputs the data to JSON.

Three other notes:
i) Make sure the docs are updated when you update code (e.g., you removed the JSON export but the docs still say the output is to a JSON file)
ii) If you are inserting the eos dict into the db, what is the procedure for someone to match that data to a particular material? e.g. I have 10,000 eos entries, how do I match an eos data to a particular material so that I can (say) grab the band structure and eos for a particular material? Normally there is a link, for instance, to a task_id that created the eos and (perhaps) a structure so one can match structure to eos.
iii) This is maybe not you, but the method to do the eos fit can simply have a reasonable default and allow the user to switch it as an optional parameter. If you don't want to do this, just add a TODO tag

computron:
@albalu I am pulling this for now, but we may want to make some fundamental changes to the way some of these workflows are done

albalu:
Thanks @computron; yes it is in my to-do list to modify it the way we discussed.

computron:
note: this PR was reverted due to a failing unit test and desire to work more on the feature

PR #81, opened 2017-02-27
Adsorption fix for bug related to removing cwd

montoyjh:
Importing matplotlib, which happens in some pymatgen modules, causes a crash of unittests because the importing matplotlib (which happens in some pymatgen modules) invokes `os.getcwd()` and some unittests remove cwds using shutil.rmtree without changing to another existent directory.  This adds os.chdir(module_dir) to all of the unittests which have shutil.rmtree in their tearDown procedure to ensure that the test is being conducted in an existing directory.


computron:
Thanks for fixing!

shyuep:
People should properly use monty.os.cd when doing such things, or monty.tempfile.ScratchDir. Those always ensure that you switch back to an existing directory. Amateurs should not use os.chdir unless they know what they are doing.

PR #84, opened 2017-03-16
NEB wflow

HanmeiTang:
## Summary

Created NEB workflow modules.

* Added workflow for CI-NEB calculations.
* The users can launch this workflow by providing (1) the parent structure, or (2) two end-point structures, and/or (3) a series of image structures.
* Support step-by-step NEB calculations, especially for those slow-converging calculations.
* Complete unit tests that cover all input options.

## TODO (if any)

Apply IDPP algo for improving initial NEB path, and some more functions.

* IDPP is a robust path-finding method used to generate a better initial NEB path. Will combine this with class **WriteNEBFromEndpoints**.
* Currently no module supports db insertion, will be available soon.
* Further analysis modules will be added.


atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -292,3 +295,77 @@ def run_task(self, fw_spec):
                                "eigenvecs": normalmode_eigenvecs.tolist(),
                                "norms": normalmode_norms.tolist()}
         return FWAction(mod_spec=[{'_set': {'normalmodes': normalmode_dict}}])
+
+
+@explicit_serialize
+class TransferNEBTask(FiretaskBase):
```
shyuep: I don't think this needs to be in glue_tasks. All NEB firetasks can be consolidated under a NEB subpackage for now. Eventually, a TransferDirectory task should be written that does a full transfer of a directory, rather than having separate Transfer* for different types of run.

atomate/vasp/firetasks/run_calc.py
```diff
@@ -147,6 +148,11 @@ def run_task(self, fw_spec):
         elif job_type == "full_opt_run":
             jobs = VaspJob.full_opt_run(vasp_cmd, auto_npar=auto_npar, ediffg=self.get("ediffg"),
                                         max_steps=5, half_kpts_first_relax=False)
+        elif job_type == "neb":
```
shyuep: I am not sure why NEB needs a specific mod. You should really leave things like half_kpts and auto_gamma alone. The user sets what he wants.

atomate/vasp/workflows/base/neb.py
```diff
@@ -0,0 +1,355 @@
...
+        n_images = len(images) - 2
+        s["_queueadapter"].update({"nnodes": str(n_images)})
+
+    if wf_name is not None:
+        s["wf_name"] = wf_name
+
+    return s
+
+
+# TODO: Enable setting using >>my_vasp_cmd<<
```
shyuep: I think these are duplicates? Think carefully about whether you need a new method or you can modify an existing one.

atomate/vasp/workflows/base/library/neb.yaml
```diff
@@ -0,0 +1,53 @@
+# A standard CI-NEB calculation
```
shyuep: Good job with this. But where do I specify # of images for the CINEB?

shyuep: Also, adding an IDPP step should be an option too in image construction.

HanmeiTang: There are several options to set up image numbers:
  1. Use yaml file: set "IMAGES" tag for NEB firework user_inar_settings (**wf_nudged_elastic_band** method in atomate.vasp.workflow.presets.core)
  2. Or it is the same to set user_incar_settings for every step in **get_wf_neb_from_structure**, **get_wf_neb_from_endpoints** or **get_wf_neb_from_images** (atomate.vasp.workflow.base.neb)
  3. If above settings are not given, image number will be determined by ```IMAGES = endpoint_distance / image_distance```. Default is ```image_dist = 0.7``` angstrom.

After this, the node number to run workflow will be updated from 1 to image number.

HanmeiTang: Thanks, I will add IDPP soon.

atomate/vasp/firetasks/glue_tasks.py
HanmeiTang: Currently, all NEB-related tasks are in atomate.vasp.firetasks.neb_tasks. 
It is possible to combine the two tasks but ** TransferNEBTask** is a little bit different:
1) NEB calculation has subfolders
2) If the number of images is not provided by users, this firetask will calculate and pass the value to next firework (NEBFW) and modify "nnodes" accordingly.

atomate/vasp/firetasks/run_calc.py
HanmeiTang: The job_type "neb" will call **VaspNEBJob**, which is used to handle the different file structures (for example, POSCARs are nested in subfolders).
Besides, all custodian arguments can be modified by the user now.

atomate/vasp/workflows/base/neb.py
HanmeiTang: Yes, it is. I already removed this "TODO".

atomate/vasp/workflows/base/library/neb.yaml
HanmeiTang: IDPP has been added to **WriteNEBFromEndpoints** (atomate.vasp.firetasks.neb_tasks) and set to default.

computron:
@HanmeiTang Can you please send me a message when this PR is ready? At the moment it seems you are still working on it.

If you do not plan for the code to be 100% ready to merge in 5 days time, please close this PR and resubmit a PR later on.

HanmeiTang:
@computron This PR is ready to merge now.

atomate/vasp/firetasks/run_calc.py
computron: Note: prior to the NEB commits, half_kpoint_first was hardcoded to False for the other custodian types. I am pretty sure that at some time, this parameter was defaulted to True in custodian (which I did not like since I am a big fan of principle of least surprise and "do no harm" for default parameters) and so in atomate I defaulted it to False so there were no surprises of a user's kpoint mesh changing. Since then, custodian now has half_kpoint_first = False as the default and there is no reason to enforce this in atomate. It is for the same reason that atomate sets auto_npar to False since for >1 node jobs this parameter can lead to unintended and confusing behavior when the user is trying to set NPAR themselves (as we saw from some user reports) yet never explicitly told custodian to change their NPAR.

For the moment, @HanmeiTang should make two changes. (i) document all new parameters as optional parameters and (ii) make sure the new optional parameters get used for all job types not just NEB (with the same defaults as before).

In the future, I will probably try to clean up some of this with a more **kwargs style approach so there isn't a need to add a new parameter to atomate every time there is a new custodian option. The only thing atomate may want to do is to deflect some of the default behavior of custodian to be more in line with atomate's principles for how workflows should operate by default.

atomate/vasp/workflows/base/neb.py
```diff
@@ -0,0 +1,359 @@
...
+from fireworks.core.firework import Workflow
+
+from atomate.vasp.fireworks.core import NEBFW, NEBRelaxationFW
+from atomate.utils.utils import get_logger
+
+__author__ = "Hanmei Tang, Iek-Heng Chu"
+__email__ = 'hat003@eng.ucsd.edu, ihchu@eng.ucsd.edu'
+
+logger = get_logger(__name__)
+spec_orig = {"wf_name": "CINEB",
```
computron: I am not sure what this is? It seems to have a lot of code that is (i) tailored for your cluster and (ii) anyway, has nothing to do with the materials science problemlike ppn or _queueadapter.

atomate/vasp/workflows/base/neb.py
```diff
@@ -0,0 +1,359 @@
...
+    # Append Firework links.
+    fws = rlx_fws + neb_fws
+    if neb_round > 1:
+        for r in range(1, neb_round):
+            links[neb_fws[r - 1]] = [neb_fws[r]]
+    workflow = Workflow(fws, links_dict=links, name=wf_name)
+    return workflow
+
+
+def get_wf_neb_from_endpoints(user_incar_settings, parent, endpoints=None, is_optimized=True,
```
computron: I don't understand the parameters here (parent and endpoints). What is parent?

This may be a misunderstanding on my part for NEB but I would think all you need to run an NEB (in the most automatic formulation) are two endpoint structures. Then the workflow can interpolate between those two endpoints and run an NEB. So I can understand a need for two endpoint structures but I don't know what parent refers to.

atomate/vasp/workflows/presets/core.py
```diff
@@ -346,3 +349,114 @@ def wf_thermal_expansion(structure, c=None):
         wf = add_wf_metadata(wf, structure)
 
     return wf
+
+
+def wf_nudged_elastic_band(structures, parent, c=None):
```
computron: see my previous comment about "parent" and about how, in my understanding, a fully automatic NEB would only require two endpoint structures.

atomate/vasp/firetasks/run_calc.py
computron: (so for now, please make sure your optional arguments (i) work for all custodian modes and (ii) are documented)

HanmeiTang: Using a half KPOINTS is based on our experience in NEB calculations, and we would claim that most of them need to do so to achieve a reasonable convergence speed. To keep the consistency between the settings in atomate, we have modified **VaspNEBJob** arguments and now only "auto_npar" and "gamma_vasp_cmd" are allowed to change. So we no longer have conflict between the optional parameters. Alternatively, we have added two optional arguments user_kpoints_settings and additional_cust_args to every fireworks, so that the user can change k-mesh and custodian settings based on their own preference.

atomate/vasp/workflows/base/neb.py
HanmeiTang: Yes, this part was not good enough. Now it is renamed as "default_spec" and used to pass certain parameters like "sort_tol" and "interpolation_type", which are related to the construction of initial NEB migration path. As the actual NEB calculations often require multiple nodes, we still need to update the number of requested nodes in spec (via "_queueadapter"). To test the performance of the current NEB workflow, we have carried out short-time tests on both SLURM and PBS queueing systems. The results are as what we expect.

atomate/vasp/workflows/base/neb.py
HanmeiTang: "parent" refers to as the perfect crystal structure (no impurities/defects) of interest. We completely agree that the NEB workflow can always start with two given endpoint structures. However, it will be used to label a group of NEB calculation (different paths) within the same parent structure in the future, which is a helpful tag in database insertion and query.

atomate/vasp/workflows/presets/core.py
HanmeiTang: Please see my response above.

atomate/vasp/firetasks/run_calc.py
shyuep: @computron Re principle of do no harm, custodian has always defaulted half kpts to False. I don't remember it ever being True by default. There may be other parameters that were set that way, but half_kpts was not one of them.

In any case, many such arguments go out of the window with something like Custodian and atomate, which basically changes parameters to fix things and prescribes a "good" approach to doing a set of calculations respectively. Our NEB workflow is a prescription for an *efficient* and *accurate* NEB calculation. In our opinion, most other people are doing it badly or inefficiently. They can override the defaults we set, but it is what the doctor recommended.

atomate/vasp/workflows/base/library/neb.yaml
shyuep: @HanmeiTang If pymatgen-diffusion is now a dependency, you will need to add it in requirements.txt

HanmeiTang: @shyuep Yes, we already added "pymatgen_diffusion==0.3.0" to requirements.txt.

HanmeiTang:
@computron The PR is ready to merge.

atomate/vasp/firetasks/run_calc.py
computron: @shyuep you are correct, half_kpts was never defaulted to True, it must have been a different parameter I was thinking of.

As for the "out the window", there is a bit of subtlety here:
- for things like the automatic preset workflows, the spirit is more like what you said - i.e., just do whatever recipe you think is best. If that means half_kpts, so be it - these are details the user doesn't necessarily want to worry about.
- however, the original intention of RunVaspCustodian (the Firework object) was to be used as a drop-in replacement for RunVaspDirect. If one used the latter, VASP would be directly called. If you did a drop-in replacement of RunVaspDirect->RunVaspCustodian, the user expectation (from my point of view) is that everything should behave essentially the same as running VASP directly except you would get error correction for free. However this only really pertains to job_type="normal" - my interpretation of that is "VASP with error correction". For job types other than "normal" (e.g. NEB), this drop-in replacement thing is indeed out the window and one is free to do whatever since those not-normal job types are like mini workflows.

PR #90, opened 2017-04-04
fix to unittests for adsorption

montoyjh:
## Summary

Fixes unittest issue in adsorption workflow


computron:
@montoyjh sorry but I think you can do better...even HeIr4 would pass the test

computron:
Btw, try testing on the alphabetical formula ... it is insensitive to how the pretty formula rules are laid out (which btw can change)

PR #93, opened 2017-04-10
Updates to elastic tensor workflow

montoyjh:
## Updates to elastic tensor workflow

* Updates ElasticTensorToDb task to use new property dictionary functionality in new pymatgen
* Updates unit test
* Updates requirements.txt


computron:
Thanks!

PR #96, opened 2017-04-20
Neb wflow

HanmeiTang:
## Summary

An additional wall time setting has been enabled for NEB calculations with some adjustments for new pymatgen and pymatgen-diffusion features.

* NEB calculations walltime settings are usually very different from the first two relaxations for parent and endpoint structures. By adding ```neb_walltime``` to the config file (neb.yaml), the user will be able to change walltime settings for NEB calculations. The default fireworks configurations is used if ```neb_walltime``` is not given.
* Change default ```user_kpoints_settings```, which is an adjustment to recent pymatgen update.
* Change ```idpp_species``` to a list of string, which is an adjustment to recent pymatgen-diffusion update.
* Minor bug fix: endpoints were not updated to ```fw_spec``` properly and this has been fixed in the commit.

## TODO (if any)

To combine NEB workflow with pathfinder feature in pymatgen-diffusion.


computron:
Thanks!

PR #103, opened 2017-05-15
New test for wf_bulk_modulus with fewer files and adding a few tools to facilitate that

albalu:
## Summary

After changing 2 default values in the wf_bulk_modulus itself I wrote a test for this workflow where the structure optimization and 2 other deformations are tested. For other deformations, as discussed, no VASP is run not even fake vasp. This is to minimize the number and size of the test files. I managed to reduce the original 75 files/folders (3.2 MB) to 43 files/folders (1.7 MB). This can be 35 (1.32 MB) if I only test one deformation but I think 1.7 MB is ok. For doing this type of test, I had to add use_no_vasp powerup and JsonToDbTask. However, the same tools can be used for other workflow tests (e.g. elastic constant)

use_no_vasp works similar to use_fake_vasp: it just replaces the RunVasp tasks with RunNoVasp (new) and VaspToDb tasks with JsonToDbTask. JsonToDbTask just takes a json file (task_doc.json) from calc_dir and pass it to tasks database.

## TODO (if any)
I am not sure if I pick the most relevant names for the new function and class.


computron:
hi @albalu 

before I review this change, it looks like you reverted some of my refactorings. e.g., eos_dict->summary_dict is now changed back to eos_dict. Can you retain the refactorings from before and then I'll look at these new changes?

albalu:
oh ok, sorry I thought they are left from before I change FitEquationOfStateTask

atomate/vasp/workflows/tests/test_bulk_modulus_workflow.py
```diff
@@ -0,0 +1,248 @@
...
+        fake_vasp_ref_dirs = {}
+        for i in range(2, ndeformations+2):
+            if os.path.exists(os.path.join(reference_dir, str(i), "inputs")):
+                fake_vasp_ref_dirs["bulk_modulus deformation {}".format(i-2)] = os.path.join(reference_dir, str(i))
+            else:
+                no_vasp_ref_dirs["bulk_modulus deformation {}".format(i-2)] = os.path.join(reference_dir, str(i))
+
+        fake_vasp_ref_dirs["structure optimization"] =  os.path.join(reference_dir, "1")
+
+        print("here")
```
computron: @albalu I think you can remove these debug prints, no?

albalu:
@computron , I reverted some of the changes I made to FitEquationOfStateTask. The latest version that I see on Github does not pass anything to db so I did not revert those changes as without the db part I would need to also change my test to be compatible. Is there a reason to_db part was omitted?

atomate/vasp/workflows/tests/test_bulk_modulus_workflow.py
shyuep: Quick comment: I strongly discourage the removal of prints, but rather use logging.debug to log debug messages. These are helpful when you need to find out what went wrong and you can just do a run with a logging.basicConfig(level=logging.DEBUG). 

That said, of course something like print("here") is completely useless from the debugging standpoint. The messages need to be a lot more useful.

computron:
@albalu I don't know why there is no way to put it in the DB. I actually put in a comment to Kiran that there should be a method to put it in the DB....

Basically just make sure that all your changes are making the code better, i.e., that any changes you are making versus what's currently on master are **intentional**, and then let me know when finished.

albalu:
Hi @computron , @matk86 had asked me to add the Db insertion part before so I did. I don't claim all this is perfect but is the best I could do at the moment. 

Regarding my effort to minimize the number of files in "test_bulk_modulus_wf.py": I tried to add and use various tools such as use_no_vasp powerup or JsonToDb (which directly passes a task.json file to VaspCalcDb) so that these may be applicable to other tests and we can minimize the number of files we use in tests.

All changes here are intentional and ready for your review.

computron:
Thanks! Note that I will make the filename "task.json" be a customizable parameter after pulling.

PR #105, opened 2017-05-17
Adsorption refactor and cleanup

montoyjh:
## Summary

This refactors the adsorption workflow to eliminate much of the code duplication that was present previously.  Also simplified implementation considerably.

* get_adsorption_wflow and get_adsorption_wflow_from_slab are merged into get_surface_wf
* get_wf_surface takes slab input, but may also be constructed to modify constructed slabs based on OptimizeFWs that relax the precursor bulk structure.
* addition of molecular optimization may be added with the add_molecule_in_box flag

## TODO
* automated analysis framework


computron:
Thanks!

PR #106, opened 2017-05-18
Hopefully fixed the ToDBTask UnitTest once and for all

shyamd:
Other tests were leaving task/counter collections that caused problems with the unit test
Fixed by clearing the whole test DB to ensure a clean run before starting the unit test


computron:
@shyamd do you know what tests are leaving things behind?

Issue #108, opened 2017-05-22
Elastic task fails with KeyError: "calc_locs" in CopyVaspOutputs


Atraii:
## System

* Version: "master"
* Python version: 3.6.1
* OS version: Fedora 25

## Summary

* Elastic task started via presets/core/wf_elastic_constant fails
* Seems to only occur if "optimize_stucture" is set to False

## Example code

```python
from fireworks import LaunchPad, FWorker
from fireworks.core.rocket_launcher import rapidfire
#from atomate.vasp.workflows.presets.core import wf_elastic_constant_minimal
from atomate.vasp.workflows.presets.core import wf_elastic_constant
from atomate.vasp.powerups import add_modify_incar
from pymatgen.io.vasp.inputs import Poscar
import os

db_dir = os.path.dirname("/data/icurtis/atomate/config/")
lp = LaunchPad.from_file(os.path.join(db_dir, "my_launchpad.yaml"))
poscar = Poscar.from_file("POSCAR_orig")
structure = poscar.structure
c_new = {"optimize_structure": False}
#c_new.update(c or {})
wf = wf_elastic_constant(structure, c_new)
wf = add_modify_incar(wf, modify_incar_params={"incar_update":{"ENCUT": 1000,
                                                               "ALGO": "Normal",
                                                               "KPAR": 4,
                                                               "NCORE": 4}})
lp.add_wf(wf)
fworker = FWorker(env={"db_file": os.path.join(db_dir, "db.json"),
                       "vasp_cmd": "mpirun -n 32 vasp_std",
                       "scratch_dir": "/zdata/icurtis/vasp_tmp"})
rapidfire(lp, fworker=fworker)
```

## Error message

```
2017-05-21 18:23:32,996 DEBUG FW with id: 55 is unique!                            
2017-05-21 18:23:33,028 INFO Created new dir /data/icurtis/Auxetic/OtherSystems/TiS/launcher_2017-05-22-01-23-32-996363
2017-05-21 18:23:33,071 INFO Launching Rocket                                     
2017-05-21 18:23:33,114 DEBUG FW with id: 55 is unique!                                               
2017-05-21 18:23:33,117 DEBUG Created/updated Launch with launch_id: 71          
[...]

PR #109, opened 2017-05-22
Elastic fix

montoyjh:
## Summary

* Fix for elastic workflow without initial optimization, also added test to ensure that workflow runs without structure optimization.  Addresses #108 .
* Should also fix former todo where re-run initial optimizations won't necessarily be fed properly to analysis FW.


computron:
Thanks!

PR #110, opened 2017-05-23
Fix TestToDB under python2 and cleanup of MongoDB collections from testing

shyamd:
Python2 seems to complain without an __init__ for the test drone. Don't know why.


computron:
Thanks!

PR #113, opened 2017-06-03
Documentation

bocklund:
## Summary

* Add running workflows tutorial to walk a new user through running a preset style workflow from a YAML file in atwf and doing the same in Python. Tutorial users also query their database for results and plot up some points in Python.

* Creating workflows guide to discuss the steps to create a workflow in Python from existing Fireworks and tasks, discuss a specific example. Also address the different machinery features at the workflow level like PassCalcLocs, env_chk, and powerups

* Add a short reference to YAML workflow files in the context of atomate


computron:
Thanks!

PR #75, opened 2017-02-13
Gibbs: phonon workflow test, results to database, improved defaults

bocklund:
# **Do not merge yet!**

I need some help in the Gibbs phonon unit test. I modeled it similarly to the existing elastic workflow unit test. In isolation, it works well, passes, and tests what I think it should test. But I get several errors when I try to test the whole atomate test suite. 

I've done a git bisect and determined that the whole test suite works up until commit 83b525d, where I add in the phonon workflow test. None of the errors are in the phonon workflow, but are elsewhere. See the comments below for more details on these. 

Once this is fixed, we can merge.

## Summary

A unit test was written for the Gibbs workflow with Phonopy. Debye mode files are included, workflow unit test coming soon. 

The results from the Gibbs workflows (information about the structure, the free energies, temperatures, and meta data) are written to a "gibbs" collection. (With this, a rename of GibbsFreeEnergyTask to GibbsFree

The following defaults for the Gibbs workflows were changed or updated

* `natoms` in `gibbs_summary_dict` changed to `nsites` to better conform with the style elsewhere
* Gibbs deformations changes
  - Default number of deformations to 11, which should be plenty for an average user of the Debye method and the most anyone using Phonopy should need. See http://dx.doi.org/10.1016/j.commatsci.2009.12.006 for more
  - The deformation matrix was changed so that the value passed from `np.linspace` represents the actual volume change. E.g. `np.linspace(-0.10, 0.10, 11)` makes 11 deformations from -10% volume to +10% volume.
* Fix phonopy `t_max` by adding several steps to the maximum temperature given to `PhonopyQHA` because it truncates several values in the fitting. Now the final `t_max` matches the input from the user.
* Use higher accuracy force INCAR settings by default (`LREAL=False`, `ADDGRID=True`)

More details for several of these can be found in the commits. 

## TODO (if any)

* Write a workflow unit test for the "debye_model" mode for the Gibbs free energy preset.

@matk86 should take a look at the changes in the Gibbs preset.

PR #114, opened 2017-06-05
Ferroelectric workflow (3rd time's the charm!)

blondegeek:
WARNING: Tests for get_wf_ferroelectric are included but are not passing. None of the other workflow tests seem to be currently passing either so I assume something nontrivial changed. **Please advise on what changes should be made.**

This pull requests adds the following classes and functions for the workflow for calculating the polarization of candidate ferroelectrics:
atomate.vasp.workflows.base.ferroelectric.get_wf_ferroelectric
atomate.vasp.fireworks.core.LcalcpolFW
atomate.vasp.fireworks.core.StaticInterpolateFW
(Some comments are included with suggestions for improving this FW)
atomate.vasp.firetasks.glue_tasks.GetInterpolatedPOSCAR
atomate.common.firetasks.glue_tasks.CreateFolder
atomate.common.firetasks.glue_tasks.GrabFilesFromCalcLoc
atomate.vasp.firetasks.parse_outputs.PolarizationToDbTask


computron:
Hi @blondegeek

All the tests should pass. Actually this is auto-checked after every commit via CircleCI which shows the state of every build.

Most likely you need to update your atomate dependencies (pymatgen, etc) via a fresh "python setup.py develop" or similar command

blondegeek:
Hi @computron 

I updated all my dependencies (which did fix several of the errors I was getting) and even tried making a fresh virtual environment. However, out of all the tests in atomate/vasp/workflows/tests directory I can only get test_vasp_workflows.py to pass. The errors are each very different (I can provide more detail if helpful).

I am also getting some very strange errors from the test_ferroelectric_workflow.py unit tests which were previously passing back when I made the first ferroelectric workflow pull request.

For example, line 165 in test_ferroelectric_workflow.py will throw an error because the workflow fireworks are empty (and do not have a "_task" keyword) unless I have a `print(cls.wf.fws)` or `print(self.wf.fws)` (depending on where you put it) somewhere (for example, around line 56) before that line runs. *facepalm at print statement sensitive executions*

```
  File "test_ferroelectric_workflow.py", line 166, in test_wf
    for fw in self.wf.fws if "polarization" in fw.name and "interpolation" in fw.name]
KeyError: u'_tasks'
```

Also, when I do have the print statement included, the tests run but then there's a failure due to a mismatch of KPOINT files. 

```
ValueError: KPOINT files are inconsistent! Paths are:
/Users/tesssmidt/Dropbox/code_ferroelectric_commit/env/src/atomate/atomate/vasp/workflows/tests/scratch/launcher_2017-06-06-06-46-12-291067
/Users/tesssmidt/Dropbox/code_ferroelectric_commit/env/src/atomate/atomate/vasp/test_files/ferroelectric_wf/interpolation_1_polarization/inputs with kpoints Automatic kpoint scheme
0
Gamma
7 7 6
 and Automatic kpoint scheme
0
Monkhorst
8 8 8
```

There's nothing inherently wrong with this error, especially since it seems the pymatgen 4.7.6 has different kpoint sampling defaults. However, if (with DEBUG=True in the previous step) I do the following:

```
rm .../atomate/atomate/vasp/test_files/ferroelectric_wf/interpolation_1_polarization/inputs/KPOINTS
mv .../atomate/atomate/vasp/workflows/tests/scratch/launcher_2017-06-06-06-40-41-501031/polarization/KPOINTS .../atomate/atomate/vasp/test_files/ferroelectric_wf/interpolation_1_polarization/inputs/.
```

I get the following error:

```
ValueError: KPOINT files are inconsistent! Paths are:
.../atomate/atomate/vasp/workflows/tests/scratch/launcher_2017-06-06-06-40-41-501031/polarization
.../atomate/atomate/vasp/test_files/ferroelectric_wf/interpolation_1_polarization/inputs with kpoints Automatic kpoint scheme
0
Monkhorst
8 8 8
 and Automatic kpoint scheme
0
Gamma
7 7 6
```

The KPOINTS files have swapped! Grrrr.....

At this point, I am tempted to throw my computer across the room to see if that will help in the debugging process. I would very much like to finalize this pull request in a timely manner, but I am baffled as to what is going on!

Can you reproduce these errors?

Tess

bocklund:
Github is saying that your glue_tasks conflicts with the hackingmaterials:master branch. And your ferroelectric_2 is reporting 2 commits head, 36 behind hackingmaterials:master.  Maybe you are developing in the middle of some broken state?

Do the tests pass if you just run them on the hackingmaterials:master branch?

bocklund:
Also with regards to print statement dependency, I think I was experiencing something similar in the Gibbs tests that I wrote in #75. I tried more after that, but never got to the bottom of it.

My test was, relatively speaking, longer running (about a minute). My best guess is that nose (which is what I use, but in pure Python I had the same issues) runs setup/teardown in parallel. Meaning files and the database, could be getting cleaned out while other tests are still running and using them. I wasn't able to find clear documentation on this and I never got around to actually reading the source code for test runners to confirm. But it seemed consistent. 

The fact that just a print statement is causing your success or failure has me questioning that conclusion, but hopefully we can figure it out and solve both problems.

Perhaps mocking the database would be better than actually accessing it. That seems to be better practice around web-dev people and would isolate each test in its own sandbox properly.

computron:
@blondegeek you might try a fresh install? @bocklund seems to have gotten it to work out OK that way - minus the ferroelectric test which looks like a legitimate bug

computron:
@blondegeek do you want to schedule a call to try to sort this out?

blondegeek:
I'll follow up via email.

computron:
Hi @blondegeek I am closing this pull request since based on last conversation it is not ready. Please re-open when ready

PR #118, opened 2017-06-06
Fix travis and misc cleanup of tests

matk86:
* Testing with python 2.7 fails on travis(3 is fine) due to numpy/scipy installation failure caused by some lapack issues. So i set up travis with conda(nothing fancy, straight up copied sample config from conda doc with minor customizations) for both 2.7 and 3.6
* All tests except one pass for both 2.7 and 3.6. The only test that keep failing is the bulk modulus wflow test(some pymongo issue on travis) but it passes on my machine(both 2.7 and 3.6).  @albalu could you take a look at the bulk modulus tests?  Also i noticed that in the bulk modulus test some of the refernce files are getting modified?!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,11 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION --file requirements.txt
   - source activate test-environment
-  - pip install -r requirements.txt
   - pip install coverage
   - pip install python-coveralls
+  - pip install nose  
 before_script:
   - source activate test-environment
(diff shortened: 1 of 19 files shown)


albalu:
@matk86 , I might know what's going on here. First to answer your question, the reference files are the task.json which are used to be directly read and inserted to task collection and their name needs to be modified for each time a new bulk modulus workflow is called.

As for the error, I am still trying to reproduce it locally but everything runs ok! However, I think the problem might be the fact that in JsonToDbTask I am trying to insert a dict from file that already contains "_id" key?  Maybe try changing the end JsonToDbTask to?:

            try:
                del (task_doc["_id"])
            except:
                pass
            mmdb = VaspCalcDb.from_db_file(db_file, admin=True)
            mmdb.insert(task_doc)

matk86:
fixed the issues. can be merged now.

computron:
Thanks!

PR #119, opened 2017-06-06
Fix docs by passing in the custom input set.

bocklund:
Also fix indentation problems.

@computron you mentioned that you pushed the docs, but it doesn't look like the new release is up.


computron:
Strange .. I just tried releasing again (sometimes it takes 15 mins to refresh, usually not longer than that)

bocklund:
@computron looks good now. Thanks!

PR #122, opened 2017-06-08
Refactor deformations wflow

matk86:
* following @computron 's suggestion, improve modularity by enabling chaining of workflows(in this case chain structure optimization with wflows that utlize deformations wflow)
* @montoyjh since you introduced the optimize_structure option(replaced in this pr), please review this pr.


bocklund:
I think `vasp_input_set` should still be an argument that controls the deformation calculations now.

atomate/vasp/workflows/presets/core.py
```diff
@@ -281,14 +297,33 @@ def wf_gibbs_free_energy(structure, c=None):
     poisson = c.get("POISSON", 0.25)
     metadata = c.get("METADATA", None)
 
-    # 21 deformed structures
+    # 21 deformed structures: from -10% to +10%
     deformations = [(np.identity(3)*(1+x)).tolist() for x in np.linspace(-0.1, 0.1, 21)]
```
bocklund: These do not deform +/- 10%, but instead +33.1% and -27.% (1.1**3 and 0.9**3) because you are applying a 10% change to each lattice vector

bocklund: You need to take the cube root of (1+x) to fix

atomate/vasp/workflows/presets/core.py
```diff
@@ -357,9 +410,26 @@ def wf_thermal_expansion(structure, c=None):
     # 10 deformations
     deformations = [(np.identity(3)*(1+x)).tolist() for x in np.linspace(-0.1, 0.1, 10)]
```
bocklund: I suggest using an odd number here to capture the equilibrium volume

atomate/vasp/workflows/presets/core.py
```diff
@@ -201,18 +202,33 @@ def wf_elastic_constant(structure, c=None):
     c = c or {}
     vasp_cmd = c.get("VASP_CMD", VASP_CMD)
     db_file = c.get("DB_FILE", DB_FILE)
-    # @montoyj - any comments?
+    # @montoyj - any comments on the following being configurable params?
```
montoyjh: Yeah, I'll remove these in my next PR, which basically has a single wf for the "MP" elastic procedure and a "minimal" one as well.

PR #129, opened 2017-06-20
Fixed dieletric_constant_no_opt

shyamd:
Previously yaml definition didn't include copy_vasp_outputs which is set by default to True in the LepsFW. This just ensures it gets set to false since there is no previous firework.


computron:
Thanks!

PR #130, opened 2017-06-22
refactor & cleanup vaspdrone

matk86:
* now its possible to skip the dos and bandstructure parsing alltogether(@tschaume might like to use this feature). Note: there is a difference  between what is meant by dos parsing by the drone(sets 'dos' field) vs dos parsing by vasprun(for bs parsing, the dos must be parsed by vasprun so that the fermi energy can be set(required by Vasprun.get_bandstructure method)).
* add 'missing' field to the 'schema' field in the task doc to. This is basically a schema sanity check report.
* added/updated tests
* some minor cleanup/ docstring improvements


computron:
Hi @matk86 

Can you explain the changes to default bandstructure mode? e.g., if I run the drone on a structure optimization run with fully default parameters for VaspDrone, how is the database entry changed from before? Or any other changes worth describing.

Also why introduce the complication of True="uniform"?

matk86:
'bandstructure_mode' now support 3 options:
1. False --> bandstructure not parsed at all (before False meant the same as parsing in uniform mode)
2. "line" --> line_mode = True in Vasprun.get_band_structure
3. "uniform" -->  line_mode = False in Vasprun.get_band_structure

why True="uniform":
1. ensure backward compatibility. Before, the default setting(bandstructure_mode=False) actually meant parsing banstructure using Vasprun without parsing the projected eigen values. I made it more explicit this time, with 'False' meaning no parsing at all.

2. ensure that bandstructure_mode can be set to either bool or string type, with 'True'/"uniform" being the default as before(no parsing of the projected eigen values). Before 'True' meant parsing projected eigen values and False meant uniform mode. In this PR, False ==> no parsing at all, True ==> it is parsed(just the eigen values) and for project eigen values one has to make that explicit by specifying bandstructure_mode = "line"

computron:
This is all very confusing and I don't understand. For example why can't I parse projected eigenvalues in a uniform band structure

matk86:
yes you are right, projected eigen values must be parsed for both 'uniform' and 'line' modes. Sorry for the mixup. pushed the fix

computron:
So as far as I can understand:

1. False -> no one will ever use this (?)
2. True -> parse basic band structure to get things like gap but don't store the full band structure in GridFS or do projected band structures (e.g., used in structure optimization and static runs)
3. "uniform" -> same as True, but also will store full band structure in GridFS and projected band structure (e.g., used in high k-mesh uniform grid runs)
4. "line" -> same as uniform, but will produce a BSSymmLine object instead (e.g., used in symm line 
band structure calcs)

To me this feels like two parameters rather than 1 parameter with 4 options of two different types.

matk86:
1. False: someone like patrick might. The main reason i did the refactor was because he wanted more control over dos and bs parsing.
2. True: parse eigen values and projected eigen values (but no BandStructureSymmLine) + insert them
3. uniform : same as true
4. line: same as unform + BandStructureSymmLine

yes i agree, i considered adding 'parse_bs' param in addition to 'bandstructure_mode' but decided against it so as to minimize changes in dependent firetasks.

computron:
What is the equivalent setting for today's structure optimization runs?

1. Parse band structure (e.g., band gap, etc) - e.g. ["output"]["band gap"] exists.
2. Do not parse projected band structure
3. Do not insert band structure into GridFS
(Do not pass Go; do not collect $200)

As far as I can tell, in your options.
False -> will produce NO output.band_gap field
True, "uniform" -> inserts into GridFS and parses projections
"line" -> invalid

matk86:
1. VaspToDb(bandstructure_mode="uniform" or "line" or True)
2. projected eigen values are parsed for both "uniform" and "line", no way to control the parsing of just the projected eigen values as of now.
3. parsing and insertion go hand in hand, no parsing  ==> no insertion,  parsing ==> insertion

i dont see why "line" is invalid? 

Anyway, it seems like this PR is too confusing. I'm ok with not merging this if you think its going to confuse the users.

computron:
I am talking about *structure optimization* runs - look at OptimizeFW in the initial code for example.  For those, I want to do a basic parse of the band structure (e.g., so I can get an approximate band gap) but I don't want to insert the band structure into GridFS or take the time to do a projected band structure (that is the current setting).

In the old code, if I set bandstructure_mode=False, it still parse the band structure and gave me a band gap in the output dictionary. But it didn't insert anything into GridFS or bother with projected band structure.

In the new code, how would I do this? See my comments above. e.g., if I set bandstructure_mode=False in the new code, it no longer gives me band gap in the output dictionary. And if I set it to anything else, it starts inserting the full band structure into GridFS

computron:
To see what I mean, perhaps just take an example of VASP output files from a structure optimization run. Then inspect the task document and GridFS collection items you get with "bandstructure_mode=False" in the old code (you won't get anything in GridFS; in the task document you'll see an output.band_gap key). Then try reproducing that same behavior in your version of the code using any of your 4 available settings.

matk86:
yep, you are right. None of those options deal with the case that you just pointed out(just bandgap, no projected eigenstates and no insertion). 
I guess i will close this PR for now and will do another one if i find a better solution to deal all the cases.

computron:
Ok - note that I wasn't suggesting to give up; I'm just trying to make sure that people's workflows don't start doing different things all of a sudden if I merge the PR

matk86:
Of course, I understand

PR #131, opened 2017-07-08
DO NOT MERGE: pasting wfs

montoyjh:
## Paste wf with "glue tasks"

@matk86 and I were talking about the new design of the elastic workflow and how it's preferable to have workflows be subunits that accomplish one thing (e. g. between the elastic and optimization wfs).  In the current implementation, it's a little wonky because a user has to specify the `copy_vasp_outputs` keyword as `False` or `True` separately from connecting the workflows (and if specified incorrectly, the workflow breaks, even if the user connecting them presumably means they want the results from the first workflow passed to the second).  Kiran suggested that a functionality that connects the workflows and _adds the appropriate glue_tasks_ to the parents/children that connect the workflows would be useful in making these connections without having to specify the appropriate keywords in the workflow constructor in order to ensure that the connected workflows work properly.

I've put a rough implementation in this PR, but I wanted to get feedback about where it should go.  It might be general enough to add to Fireworks proper (e. g. to `Workflow.append_wf`), but the design principle of "glue_tasks" is something that might be appropriate to keep in atomate.  If it should stay in atomate, I'm not sure whether it should stay in utils or be somewhere else.  When that's determined, I'll write a unit test.

### DIFF
--- a/atomate/utils/utils.py
+++ b/atomate/utils/utils.py
@@ -277,3 +277,53 @@ def load_class(modulepath, classname):
     """
     mod = __import__(modulepath, globals(), locals(), [classname], 0)
     return getattr(mod, classname)
+
+
+def paste_wfs(wf_1, wf_2, parent_fw_ids=None, glue_tasks_1=[], 
+              glue_tasks_2=[], **append_kwargs):
+    """
+    Function to "paste" two workflows together and
+    optionally add glue tasks between parents and
+    children from the first and second workflow
+    respectively
+
+    Examples:
+        new_wf = paste_wfs(optimize_wf, elastic_wf,
+                           glue_tasks_1=[PassCalcLocs()],
+                           glue_tasks_2=[CopyVaspOutputs(contcar_to_poscar=True)])
+
+    Args:
+        wf_1 (Workflow): First workflow to be connected, also
+            the workflow from which to take parent_fw_ids
+        wf_2 (Workflow): Second workflow, the workflow for
+            which roots will be made children of the parents
+            specified by parent_fw_ids
+        parent_fw_ids ([int]): fw_ids to use as parents
+            from the first workflow to children corresponding
+            to root_fw_ids from the second workflow
+        glue_tasks_1 ([Firetasks] or {fw_id: [Firetasks]}):
+            firetasks to append to the end of the firetasks
+            for each firework specified by parent_fw_ids.
+            If a list, adds these firetasks to every parent.
+            If a dict, adds the fireworks specifically by keys
+        glue_tasks_2 ([Firetasks] or {fw_id: [Firetasks]}):
+            Similar to glue_tasks_1, except added to the
+            beginning of Firetasks in root_fws of wf_2
+        **append_kwargs (dict): kwargs for append_wf
+    """
+    if parent_fw_ids is None:
+        parent_fw_ids = wf_1.leaf_fw_ids
+    if child_fw_ids is None:
+        child_fw_ids = wf_1.root_fw_ids
+        
+    if isinstance(glue_tasks_1, list):
+        glue_tasks_1 = {k: glue_tasks_1 for k in parent_fw_ids}
+    if isinstance(glue_tasks_2, list):
+        glue_tasks_2 = {k: glue_tasks_2 for k in child_fw_ids}
+
+    for fw_id in parent_fw_ids:
+        wf_1.id_fw[fw_id].tasks.extend(glue_tasks_1[fw_id])
+    for fw_id in child_fw_ids:
+        wf_2.id_fw[fw_id].tasks = glue_tasks_2[fw_id] + wf_2.id_fw[fw_id].tasks
+
+    wf_1.append_wf(wf_2, parent_fw_ids, **append_kwargs)
--- a/atomate/utils/utils.py
+++ b/atomate/utils/utils.py
@@ -313,8 +313,6 @@ def paste_wfs(wf_1, wf_2, parent_fw_ids=None, glue_tasks_1=[],
     """
     if parent_fw_ids is None:
         parent_fw_ids = wf_1.leaf_fw_ids
-    if child_fw_ids is None:
-        child_fw_ids = wf_1.root_fw_ids
         
     if isinstance(glue_tasks_1, list):
         glue_tasks_1 = {k: glue_tasks_1 for k in parent_fw_ids}


bocklund:
This looks useful. What do you think about changing the API to be more like concatenate, which agrees more with other Python tools?

It would be nice then to further generalize it so that you can pass a list of workflows to concatenate.

bocklund:
In particular, having it be named 'concatenate_wfs' which is immediately clear to users to common scientific packages.

montoyjh:
Sure, but I want to settle on the basic functionality for connecting two workflows first (and specifically where it should go).

shyuep:
I don't think the term "pasting" two wfs is correct. It should be merging two workflows.

One thing to think about is whether a Workflow or FireWork should develop a connector type API. That way, you don't have to specify glue tasks explicitly. E.g., a job can state that it supports passing outputs A, B, C, D, to the next level. Another job can say I support taking in inputs from previous job. When you add the two jobs, they automatically see that they can connect based on the support input/output connectors and automatically just connect. 

As an aside, it is the opinion of this coder that Workflow in FireWorks should basically implement a collections.Sequence API. E.g. things like append_wf should really just be append. Ideally, things like "+" operators etc. should work, which would accomplish the merging.

montoyjh:
Kiran and I discussed automating the connection functionality a bit as an alternative to what I'm proposing here, but our thinking was more along the lines of having existing firetasks more or less try to figure out whether there were prior tasks relevant to the current one.  I like @shyuep's suggestion of automated connections that are essentially added in the context of joining fireworks and/or workflows, but I think it's going to require a bit more thinking/effort to implement (and it seems like it would mostly occur at the Fireworks level).

The only disadvantage I see to this is that detecting problems becomes more difficult because most of the conceivable errors are soft (e. g. when a user connects workflows without knowing these details, and then VASP outputs are copied in a case where they're not supposed to be, or calc_locs isn't properly set and VASP outputs aren't copied despite the user expecting them to be), not really a reason to avoid it, just something to keep in mind.

shyuep:
I would actually avoid thinking about this in terms of VASP per se. Right now, Fireworks already support passing data from one fw to the next. All that is needed is some special indicator for passing file locations, and another for specifying what input file locs are accepted. Each fw then just needs to match input file loc with the preceding file loc and then just do the copy. File loc are then just handled as a special form of data. 

There are a few things that ideally need to be done to make this a truly robust system:

i. Specifying which input file locs are non-optional. For example, a BS run must have a previous CHGCAR. This will at the very least prevent obvious errors where an expected file is missing and partially deal with the disadvatnage you mention.
ii. A transparent way of un-compressing and renaming. E.g., if the preceding fw has a file_out_loc: {"CHGCAR": "/path/to/CHGCAR.relax2.gz"}, then the code will automatically copy CHGCAR.relax2.gz to current loc and rename it as CHGCAR after ungzipping it. This will decouple the desired output file names and compression from the input filename and compression (or lack thereof).

The main reason to go to this system is that all fw then becomes truly encapsulated definitions of expected input and expected output of either data or files. There is no ugly adhoc glue_tasks to copy X and Y if we are gluing wf1 to wf2, but not wf2 to wf3. 

I guess this will require the BDFL of Fireworks @computron to decide if it is something he wishes to see. I think it can be done in a way that is backwards compatible.

computron:
Ok there is a lot of discussion here,  I will try to cover everything:

* @shyuep says; " Workflow in FireWorks should basically implement a collections...Ideally, things like "+" operators etc. should work, which would accomplish the merging."

Note that Workflows are not sequences but DAGs, more like trees. For example, a typical workflow connection we do is starting with this workflow:

```
  structure_opt
       |
    static run
    /      \
uniform    line
```

and "append" something else to the uniform run. But the user must specify that the append command goes to the "uniform" run and not the "line" run. Yes we could do things like throw errors if there are multiple possible leaf nodes and the user tries to perform a ```+``` operation but I'd rather just keep things explicit.

* @bocklund says: "In particular, having it be named 'concatenate_wfs' which is immediately clear to users to common scientific packages."

Basically the same thing as the above answer, it's not like string concatenation or sequence append - more like a tree - and I don't want people to have the same mental image. 

* @shyuep says: "I would actually avoid thinking about this in terms of VASP per se. Right now, Fireworks already support passing data from one fw to the next. All that is needed is some special indicator for passing file locations, and another for specifying what input file locs are accepted."

FireWorks contains a "_pass_job_info" reserved keyword that passes the job info to the next FW without the user needing to write a "glue" task for it. It is then up to the next FW to do something appropriate with it - i.e., the next FW won't just automatically copy files or anything. The next (child) FW needs some sort of CopyTask that does the appropriate thing.

For atomate, I considered just having the parents always have "_pass_job_info=True" in the spec. Then the parent doesn't need to have any kind of glue tasks and only the child FW needs to have a task that copies files or does whatever. But, I felt that we probably wanted to do more complex things with glue tasks at some point (like have some custom notation to keep track of filesystem, or to look 2 FWS upstream rather than just the immediate parent) and rather than have 2 different systems for gluing things together (FWS native _pass_job_info for simple,  glue tasks for complex data passing) I thought it better to just always have glue tasks from the beginning that one could make more or less complicated as needed.

Overall, if someone wanted to take the time to come up with a very robust and generic to application (i.e., not VASP specific) way to copy data between various Fireworks in a Workflow that's better than the simple ``_pass_job_info``, I'd certainly be for it. For inclusion in atomate to replace the glue tasks it would need to encapsulate the current use cases of the glue tasks as well as not make the workflows more confusing for new users. (Another hesitation I had about using "_pass_job_info=True" is that a user can easily forget it, whereas the glue task is nice and explicit).

* @montoyjh says: " I wanted to get feedback about where it should go." @shyuep says "I don't think the term "pasting" two wfs is correct. It should be merging two workflows."

How about adding a merge_wf method to the Workflow class in FireWorks? You won't need to take in two workflows as arguments since the first workflow is already there.

* @shyuep says: "One thing to think about is whether a Workflow or FireWork should develop a connector type API. That way, you don't have to specify glue tasks explicitly. E.g., a job can state that it supports passing outputs A, B, C, D, to the next level. Another job can say I support taking in inputs from previous job. When you add the two jobs, they automatically see that they can connect based on the support input/output connectors and automatically just connect."

I would see this as a layer on top of FireWorks or perhaps some package in FireWorks. For example, one could imagine a tag in the spec that tells what kind of data a Firework emits and also what kind of data a Firework accepts. There are some workflow systems that I previously researched that work like this, e.g. one of the most directly comparable to this idea would be Makeflow which tries to use makefile-style notation for defining workflows in terms of data dependencies.

Personally I prefer the more direct method where I can really see what I am doing rather than having the system make the connections for me. To me it's easier to follow what's happening in a workflow and to debug anything that goes wrong when all the statements to set up tasks, connect them, etc. are explicitly in the code rather than handled by some underlying API. But I can see how in some cases someone would indeed like something like a "makefile" style workflow - e.g., the "structure optimization" emits an structure data object, which can be received by a "static run" that emits a "charge density" object, which is accepted by nscf runs, etc. Again I would vote that anything like that just becomes a layer on top of FireWorks

bocklund:
Since it's a tree, what about 'insert' syntax? Inserting a new workflow at index -1 would mean to append, concatenate, or merge on to the end, but we can also insert Fireworks or full Workflows into individual indices and update the downstream nodes accordingly.

shyuep:
I will only respond to the question of glue-tasks vs pass_job_info. Neither is what I was proposing, though pass_job_info *enables* it to be done. Yes, you can say well, pass_job_info will send the directory to the next Firework, and that Firework can do whatever it wants with it. To me, that is too much thinking and duplication involved. Say a static task precedes both a BS and Uniform Firework. I will have to code the specific handling of file copy (including logic of file formats, what files I should bring over, and what should not be brought over) into both the BS and Uniform Firework implementations. If for some reason the pass_job_info format gets changed, I will need to redo the implementations of BS, Uniform and many other places where this gets used.

My proposal is simply to treat files as a special data type that gets moved between Fireworks, with a proper language to specify what files are expected and provided by each FireWork. This is no different from the current input/output data, except that this file transfers data gets handled specially and transparently. I would say this solves 90% of the glue-task requirements in a much simpler framework. It is also a "glue" if you like that is used by *a lot* of use cases.

Nothing is preventing glue-tasks from continuing to exist on top of the above framework. In any case, I am a bit puzzled by the need for glue tasks in the first place. What is accomplished by glue tasks that is not accomplished by inserting an additional Firework? Let's say I have a static and BS run, and between those two runs, I want to do some data processing, what's stopping me from doing StaticFW->DataPRocessingFW->BSFW?

So basically, the file handling protocol that I propose is mainly to make Fireworks more user-friendly and simply a lot of atomate. If you have complex needs, glue tasks can continue to exist.

montoyjh:
>How about adding a merge_wf method to the Workflow class in FireWorks? You won't need to take in two workflows as arguments since the first workflow is already there.

Yeah, this was similar to my thought as well.  My approach previously was simply to add two optional keyword arguments to `Workflow.append_wf` that allowed for the addition of firetasks at the leaves of the first wf and the roots of the second.  This wouldn't disrupt any of the current usage of append_wf, just adds a bit of convenient functionality.  I'll issue a PR from that code on FireWorks, which I can move into its own function if we decide it needs to be separate from `append_wf`.

computron:
@shyuep says: "My proposal is simply to treat files as a special data type that gets moved between Fireworks, with a proper language to specify what files are expected and provided by each FireWork."

I have no objections to a proposal that improves the file passing procedures in FireWorks. But, someone needs to fully design it (e.g., the "proper language") and code it. I don't have room in my schedule or that of my group, so you would need recruit someone to do it. I've put it in my "someday" list of things to do, but honestly that is likely not going to happen.

@shyuep asks: "What is accomplished by glue tasks that is not accomplished by inserting an additional Firework?"

You could insert an additional Firework, but glue tasks give you the option of putting it in an existing FW or inserting a new Firework (e.g., with the glue task in it). I imagine most people (including myself) would prefer to put in an existing FW since a new FW means that in some launch modes (e.g., qlaunch singleshot) you would use an entire queued job to just run that one data processing step. Thus the task is better than an additional data processing FW.

@montoyjh says: " My approach previously was simply to add two optional keyword arguments to Workflow.append_wf"

Yes, after sending the comment I had the same thought as well. It's better to just add the functionality to the existing append_wf, there is no need to have two functions to append_wf.

I am closing this PR since we decided to put the function in FWS, and any further discussion of file passing functionality defaulted in FWS should really go to that repo rather than atomate.

PR #134, opened 2017-07-21
add the ability to copy additional files in StaticFW

albalu:
## Summary

I wanted to define a new workflow where WAVECAR is also copied from the previous static step. I needed to add the argument additional_files to StaticFW to enable that. This change is backwards compatible as additional_files defaults to None.

Also, I had to change CopyVaspOutputs to what I think is simpler and fixed a bug. The bug: in the previous version, if contcar_to_poscar and "CONTCAR" not in files_to_copy, it would raise an error when files_to_copy was still None (if "$ALL" in additional_files).

I also tested this via the new workflow that I defined.


shyuep:
I think you should use the new file copy features in fireworks.

PR #136, opened 2017-07-22
TOEC elastic

montoyjh:
## Refactor and expansion of elastic tensor workflow

* The elastic constant workflow has been refactored to generate deformations based on genuinely independent strains, rather than perturbing the deformation matrix directly.
* The fitting method for elastic tensors now uses the true stress (e. g. the second piola-kirchoff stress) and can accept keywords to use one of the finite-difference, independent strain linear interpolation, or direct pseudoinverse methods to fit the tensor.  For higher order tensors, finite-difference must be used.  Note that all fitting data is inserted into the final elastic tensor document, enabling uses to do further fitting and/or inspection if desired.
* The elastic workflow now supports symmetry reduction via the sym_red keyword.
* A "minimal" elastic workflow with the fewest possible deformed structures is introduced, which includes symmetry reduction.
* The elastic workflow can be used to calculate the third-order elastic constants.
* Unit tests have been revised for the new deformation scheme, the new fitting procedure, and for the third-order functionality.  Rather than including VASP test files for all of the higher-order deformations, I use a spec generated from a standard workflow in the last firetask.

## TODO

* Benchmarking of the TOEC workflow is still ongoing, and may involve further updates to default VASP parameters
* There's an issue in the insertion of task documents from higher-order elastic constants for the TOEC workflow.  These tasks require a very dense k-point mesh, and if used for elemental materials with primitive cells (for example) may have bandstructures that are too large in size to be inserted into the mongo database.  It appears that the bandstructure is included by default in the task document.  In my development code, I've just moved the code which includes the explicit bandstructure object in the task document into the conditional based on the mode (e. g. if the user selects a mode, the bandstructure is included in the task doc, if not, it is not), but I'm not sure if this is appropriate for general use.  If it is, I can add it to this PR or issue another one later.


montoyjh:
Working on fixing the travis tests.

montoyjh:
Seems the tests are passing, should be ready to merge.

computron:
Thanks!

PR #140, opened 2017-08-03
Small change to VaspDrone to pick up volumetric data metadata

mkhorton:
This simply adds information to `calcs_reversed` on what volumetric data is present for easier querying of the tasks collection.

This doesn't address *how* we may one day want to store the volumetric data itself (#139), which is a much larger question.

@tschaume is using the Atomate drone on large set of tasks soon, so this (or something like it) would be useful to have for that, to save having to re-run it later.

### DIFF
--- a/atomate/vasp/drones.py
+++ b/atomate/vasp/drones.py
@@ -319,7 +319,7 @@ class VaspDrone(AbstractDrone):
         d["output"]["is_metal"] = bs.is_metal()
         d["task"] = {"type": taskname, "name": taskname}
 
-        d["datafiles"] = self.process_raw_data(dir_name, taskname=taskname)
+        d["output_file_paths"] = self.process_raw_data(dir_name, taskname=taskname)
 
         if hasattr(vrun, "force_constants"):
             # phonon-dfpt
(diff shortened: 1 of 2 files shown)


computron:
How about WAVECAR? Even if you don't store it, it's volumetric data that was computed and may exist on the file system

shyuep:
Actually, if we are going to be only storing filepaths, the easier thing is to simply include all files. This costs very little in space. For example, I can foresee that people might want to use the PROCAR (if any). maybe re-parse the OUTCAR to find info that were missed, etc. 

Technically, since you have the directory of the run, you can always go back to find out what;'s in there. But having full file paths help in terms of just simply pulling the file via some paramiko based query.

computron:
All filenames sounds good to me. But I would use relative paths compared with the dirname that's already at the root level of the document. That way, if dirs get moved, we can just update the top level dirname and not all the individual filepaths.

mkhorton:
Just pushed an update for this, at the moment we have:

```
possible_files = ('CHGCAR', 'LOCPOT', 'AECCAR0', 'AECCAR1', 'AECCAR2',
                  'ELFCAR', 'WAVECAR', 'PROCAR', 'OPTIC')
```

Any others we want?

shyuep:
I don't think PROCAR is a volumetric data. 
Instead of having a field called volumetric data, I would just suggest d["output_file_paths"]. Include OUTCAR, etc.

mkhorton:
Yeah, I renamed it to the rather uninspiring `datafiles` in the recent change, will change it to `output_file_paths` instead

PR #144, opened 2017-08-16
Modify potcar

montoyjh:
## Summary

Functionality to modify potcars of Vasp runs much like the modify_incar functionality.

* Uses a {"element": "symbol"} dictionary to select appropriate potcars to make substitutions for.  Can also use different functional potcar sets (e. g. PBE_52, LDA_US, etc.) if specified.
* a minor update to PassResult to allow for callable attributes to to be passed as the result of the callable via the api.

## TODO (if any)

At some point, it might be prudent to have a more general function like "modify_vasp_inputs" that just merges the functionality of both these and potentially kpoints.


montoyjh:
Passing now.  Also added an option to setUp without a launchpad in AtomateTest.setUp (default is to set up launchpad).

PR #145, opened 2017-08-16
Add pretty formulas to all *ToDb tasks

bocklund:
## Summary

Currently these collections are hard to search for specific structures because they create their own collections.
These changes make it somewhat easier to find calculations within collections.

### DIFF
--- a/atomate/vasp/firetasks/parse_outputs.py
+++ b/atomate/vasp/firetasks/parse_outputs.py
@@ -181,6 +181,7 @@ class BoltztrapToDb(FiretaskBase):
         v, o = get_vasprun_outcar(bandstructure_dir, parse_eigen=False, parse_dos=False)
         structure = v.final_structure
         d["structure"] = structure.as_dict()
+        d["formula_pretty"] = structure.composition.reduced_formula
         d.update(get_meta_from_structure(structure))
 
         # add the spacegroup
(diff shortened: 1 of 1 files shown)


matk86:
on a related note, i think it might be worthwhile considering refactoring of the *todb tasks to use custom drones(similar to vaspdrone, maybe subclass it) that handle the corresponding task doc creation...also @montoyjh suggested moving those final analysis to builders(not sure where it would go though as there is a separate MP effort(maggma/emmet packages) to handle all the building)

computron:
Btw the get_meta_from_structure() function in atomate utils is intended to more generally give you a bunch of structure parameters to query on (not just pretty formula).

It could use updating, though. For example, "formula_pretty" is already there but is called "formula_reduced" (more in line with the old MPWorks nomenclature than atomate). But the better solution would be:

* update get_meta_from_structure() names
* call this method to generate more query-friendly terms from a structure

Note that we should also just reconsider how these db insertions of analysis tasks are happening in the first place. A builder based approach might be better.

bocklund:
Maybe that's what I should have changed then? 

IMO that should be inserted into every one of these tasks that go to a specific collection, so you can just query on those on a collection basis.

computron:
@bocklund if you are able in the short term to modify get_meta_from_structure() to use updated names and use that instead of your pretty_formula code, that would be better

computron:
actually let me just update the names now in get_meta_from_structure(). Then you can just update your pretty_formula code in this PR to use get_meta_from_structure

computron:
(ok done updating get_meta_from_structure())

PR #149, opened 2017-08-16
Config script

shyamd:
A simple config script for atomate. 
 * some basic defaults for most values
 * can boot-strap my_launchpad.yaml, my_fworker.yaml, and my_qadapter.yaml from existing files


bocklund:
As another side comment, @computron has suggested (and the documentation reflects) using rapidfire instead of singleshot in the queueadapters. We should provide the option for users to use a reservation + singleshot approach, but IMO the default should be consistent with our docs to launch with rapidfire within the walltime.

shyuep:
I suggest looking at pymatgen-db's `mgdb init` as an example.

shyamd:
@bocklund I agree. I've coded this up for the use cases I've experienced in on-boarding undergrads to using atomate and put the pull request up to get feedback. I think a fully interactive mode would be a good idea, but shouldn't take away from a simple noninteractive mode. Let me add that in today.

shyamd:
temp closing to clean up the pull request

PR #152, opened 2017-08-18
Install script

shyamd:
Continuation of #149 because I'm lousy at git. 

Basic setup script for new atomate environments:
- Sensible defaults
- Allows overriding with your own files but will update pertinent parameters like the directory paths for you
- interactive mode that will walkthrough all the parameters
- sticking with rlaunch singleshot by default for now since it's more robust in terms of the ways in which qlaunch can be run.

Possible TODOs for a future:
- auto detect queue type


bocklund:
I think that somewhere along the line with the preset workflows, the FWorker requires that `scratch_dir`, `incar_update`, and I think the other common env_chk params are defined, even if they are `scrach_dir: null`. 

Maybe @computron can comment on whether this behavior is intentional, but currently I think a user would not be able to run this by default. This is mentioned on the [google groups site](https://groups.google.com/d/msg/atomate/xVghmE3YOco/jPmJVKdGBwAJ)

shyamd:
Good point. I believe empty strings work for those too and have updated accordingly

computron:
I merged and decided to give it a spin but a few issues

1. directory is an optional arg but the script dies without it
2. I get really ugly serializations like: 
```
'': !!python/unicode <<FILL>>
!!python/unicode _fw_q_type: !!python/unicode SLURM
account: !!python/unicode <<FILL>>
constraint: !!python/unicode <<FILL>>
cpus_per_task: !!python/unicode <<FILL>>
gres: !!python/unicode <<FILL>>
job_name: !!python/unicode <<FILL>>
launch_dir: !!python/unicode <<FILL>>
license: !!python/unicode <<FILL>>
nodes: !!python/unicode <<FILL>>
ntasks: !!python/unicode <<FILL>>
ntasks_per_node: !!python/unicode <<FILL>>
post_rocket: !!python/unicode <<FILL>>
pre_rocket: !!python/unicode <<FILL>>
qos: !!python/unicode <<FILL>>
queue: !!python/unicode <<FILL>>
rocket_launch: !!python/unicode rlaunch -c /Users/ajain/tmp singleshot
walltime: !!python/unicode <<FILL>>
```
The files are completely unreadable!

For FWS objects, you can just use to_file(). e.g. "my_fworker.to_file("my_fworker.yaml") Hopefully this should give less ugly serializations

4. There is no fw_config.yaml, so you then you must run all the FWS commands inside the chosen directory or add a bunch of flags like ``-c`` to commands

5. Even though it says there are sensible defaults, it wouldn't be clear to a user what needs to be changed. For example, the my_launchpad.yaml looks filled in but a lot of people probably aren't running local MongoDB.

I'll leave this script in for the moment but I would currently suggest that people just follow the instructions in the online documentation.

shyuep:
For yaml, use safe_dump. That usually handles this kind of ugly crap.

computron:
I prefer just calling to_file() from the FWS object. e.g. launchpad.to_file("my_launchpad.yaml") or launchpad.to_file("my_launchpad.json"). This also makes some of the code in this method unnecessary (like build_db_json, not sure what does but looks like it should be one liner: launchpad.to_file("my_launchpad.json"))

shyuep:
Yes, of course it shouldn't duplicate stuff. But in instances where it is necessary to write other config not FW related, I am suggesting to use safe_dump. I will leave whoever is doing this to do whatever option is best.

computron:
Ok on further thought I am only going to be supporting the atomate install via the online instructions. I just can't be fielding questions or maintaining installation infrastructure other than that one place.

@shyamd I will send you a copy of the latest script and then remove it from the repo. I would suggest you add it to your personal Github account as a "gist" or something. That will allow others to modify it while making it clear that this is a separate thing.

PR #154, opened 2017-08-20
Update install docs by being more consistent with qlaunching

bocklund:
I think we had previously agreed to encourage people to not use reservation mode and use rapidfire within a qlaunch.
I thought I updated the installation guide to reflect this, but it appears I did not (or at least didn't push those changes up).
The tutorials all assume qlaunch run 'normally', without reservation mode and with rapidfire used in qlaunch (users should only be running qlaunch singleshot).


computron:
Sorry looks like we intersected on changes to the docs. Could you resolve the conflicts in "docs_rst/installation.rst" and resubmit the PR? Please try to avoid losing any of the changes I made.

The stuff in "docs" doesn't matter, we'll just recompile the docs anyway.

bocklund:
Should be fixed. We both tried to remove the automated installer.

computron:
ok thanks!

computron:
@bocklund  - note please use the invoke command "make_doc" (not "update_doc") to do the doc rebuild. Otherwise you get a confusing commit message about updating the version of atomate

PR #155, opened 2017-08-22
refactor LepsFW

matk86:
* LepsFW split into DFPTFW + RamanFW
* old LepsFW with deprecation warning retained for backward compatibility
*  DFPTFW and  RamanFW use _files_out/_files_in feature of fireworks to copy files instead of passcalcloc/copyvaspout fireteask combo


computron:
Looks like this is backwards incompatible with any prior workflows that may be stored in a database with phonon=True?

shyuep:
Note that if we are refactoring, I would much rather we use the new FireWork feature of transparent file copy rather than the CopyVaspOutputs tasks.

matk86:
cannot copy files with dot extensions using the _file_out/_file_in:

 pymongo.errors.OperationFailure: The dotted field 'vasprun.xml' in 'spec._files_out.vasprun.xml' is not valid for storage.

computron:
@shyuep can you look into this?

shyuep:
I think the reason is that you chose a name that has a ".". You just need to call it vasprunxml instead of vasprun.xml. E.g., {"vasprunxml": "path/to/vasprun.xml"}. This was the reason why there was a name mapping for both files_in and files_out. It was to get around this MongoDB restriction.

computron:
@matk86 is this ready for merge?

I'll merge this when you're ready since I don't know of anyone using the "phonon" parameter but in the future we should avoid such backward incompatibility. At the very least, it should be mentioned when submitting the PR so that when we do a release we can clearly explain it.

matk86:
yes, updated the pr comment..old lepsfw retained for backward compatibility

computron:
ok, will merge after tests pass

PR #156, opened 2017-08-22
copy piezo tensor to output if applicable

shyamd:
copy piezo tensor to output in drone
checks to ensure the structure is non-centrosymmetric
useful to move this to output for easier searching

Also considering adding a non-centrosymmetric tag to the spacegroup, but not sure if that makes sense


computron:
Thanks!

Issue #19, opened 2016-05-03
Default procedures for optimizations


shyuep:
I think the whole MP workflow can use a few tune-ups. For structure relaxations, instead of the current dumb "double-relaxation" that custodian does, I suggest the following recipe.

1. Start with looser kpt grid, e.g., 0.5 of normal grid.
2. Second run, do the actual grid we want + EDIFFG  = -0.05. This last criteria is not used in MPVaspInputSet right now, but is actually critical to get good structures.
3. Check CONTCAR volume - POSCAR volume. If volume change > 2%, repeat step 2.

I don't have time to do this myself, and I am a bit confused about where each of the many RunVasp do in the actual grand scheme of things. So I suggest someone else take the lead. Actually, the full_opt_run in Custodian already does this without the EDIFFG part. I will push something shortly which will enable this.

computron:
Two questions:

1. Why does it help to start with a looser k-point grid? I understand that the first calc will take less time. But then won't the second calc take more time since the solution won't start off in a state that is as well converged?
2. I've suggested force convergence (negative EDIFF) in the past, but Kristin (and maybe Geoffroy?) has told me that it is not completely reliable, i.e. does not always converge. Thus I have been reluctant to make force convergence the default strategy for high-throughput.

Note that ``full_opt_run`` is already supported in the same way as the double relaxation of RunVaspCustodian. Just check the docs of RunVaspCustodian. For example, if you want the band structures to use full_opt_run, just switch the job_type this one line of code:

matmethods/vasp/workflows/automatic/standard.py:23

```
wf = use_custodian(wf, fw_name_constraint="structure optimization",
                       custodian_params={"job_type": "double_relaxation_run",
                                         "max_force_threshold": 0.25})
```

shyuep:
It helps for the first relaxation because the first geometry step is usually the one that takes the most steps. For example, let's say the first relaxation is 50 steps, and the second relaxation is 3 steps. By halving the kp, you may halve or even 1/4 the time for the first relaxation. Even if the second relaxation takes a few more steps, you are ahead. In general, geometry relaxations aren't all that sensitive to the kpt grid, and the first relaxation's data is completely useless except from the standpoint of getting an initial geometry guess.

[...]

PR #162, opened 2017-10-02
New optimization scheme

shyamd:
Updating the optimization per #19 

Switched ediffg to default to -0.05
changed default to not include MaxForceErrorHandler 
add in half_kpts_first_relax and set to default to true

Currently, we use double relaxations. Would it make sense to switch to full opt since that enables the 2% vol change check @shyuep recommends? Does this primarily mitigate issues with Pulay stresses?

In reference to @montoyjh's suggestion of looking into the FFT  grid size, I ran some test calculations of GGA+U and SCAN optimizations for Si, Fe, Al2O3, Fe2O3, and GaN after rattling all atoms + one significant perturbation. In all cases setting ENAUG to 4*ENCUT was as at minimum 7% faster average over 4 tries, and in some cases as much as 54% faster. The resulting structures and energies all matched with the expected values from MP, or in the case of SCAN were self-consistent. The benefit seems to scale with the complexity of the calculation, more atoms = bigger speedup.

### DIFF
--- a/atomate/vasp/firetasks/run_calc.py
+++ b/atomate/vasp/firetasks/run_calc.py
@@ -83,11 +83,12 @@ class RunVaspCustodian(FiretaskBase):
         gamma_vasp_cmd: (str) - cmd for Gamma-optimized VASP compilation.
             Supports env_chk.
         wall_time (int): Total wall time in seconds. Activates WalltimeHandler if set.
+        half_kpts_first_relax (bool): Use half the k-points for the first relaxation
     """
     required_params = ["vasp_cmd"]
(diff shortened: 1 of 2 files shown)


shyuep:
I would say the force convergence is already be a good start. As for volume change, 2% is probably too small. I think 10% should be good enough. In any case, this affects a hopefully small minority of cases. Usually double relaxations do the trick. Might be useful to just use double first, and later, we can analyze how many have > 10% voume changes in the second step. We can then add additional relaxations as necessary then.

computron:
Thanks! Let's discuss a bit tomorrow at MP meeting before I pull

PR #163, opened 2017-10-04
Update Bandstructure parsing in Drone

shyamd:
## Summary
Fixed a bug in the drone for parsing runs with IBRION==1 where VASP fails to output efermi
Added in logic to determine if a run is uniform or line for nscf calculations and create the right type of Bandstructure class

### DIFF
--- a/atomate/vasp/drones.py
+++ b/atomate/vasp/drones.py
@@ -276,7 +276,7 @@ class VaspDrone(AbstractDrone):
         """
         vasprun_file = os.path.join(dir_name, filename)
 
-        vrun = Vasprun(vasprun_file, parse_eigen=True, parse_projected_eigen=True)
+        vrun = Vasprun(vasprun_file, parse_projected_eigen=True)
 
         d = vrun.as_dict()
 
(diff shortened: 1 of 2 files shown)


computron:
Thanks!

Can you make the following changes?

1. Change the default "Auto" to "auto". I know that the code works with any capitalization but I prefer if we stick to lower case for the default

2. For the try-except(s), if bandstructure_mode=True then it shouldn't pass on the error. In that situation it is a real error and should be treated as such

3. Document the "auto" mode in docstring

4. What is the performance penalty for doing Vasprun (w/parse eigen=T, needed later on for line 324) and then another BSVaspRun (which again parses eigenvalues) vs the old way? 

5. I don't understand the num_kpts logic for deciding between line mode and uniform mode. Can you explain that to me? Note that the KPoints object has an explicit "mode" which can be line_mode

shyamd:
1-3) done
4) parse_eigen is True by default in the vasprun.xml
I did a quick set of tests and I don't see any decrease in speed for Vasprun + BSVasprun( parse_projected_eigen=True) vs. Vasprun( parse_projected_eigen=True)
It's really only because I need some of the logic from Vasprun to decide if projections should be parsed. 
5) We don't get back the linemode style when the Kpoints object is made from a Kpoints file, so I'm just checking for the existence of explicit kpoints. There might be a better way to do this.

computron:
Ok thanks!

Regarding 5 it would be better if the style was parsed back. But I will accept it as is

Issue #165, opened 2017-10-06
TestWriteVasp unit test fails on system w/VASP_PSP_DIR set


computron:
@montoyjh 

Running ``test_write_vasp.py`` on my local system fails. This is because a few of these tests try to modify the POTCAR to ``Si_alt``. That POTCAR only exists in the atomate tests (is not a real VASP POTCAR). If the user doesn't have a VASP_PSP_DIR set on their system, it's fine since AtomateTest.setUp() will set the testing dir with Si_alt POTCAR as the VASP_PSP_DIR. But if the user does have a legitimate VASP_PSP_DIR set up, then the test fails

You should modify the test so that it either doesn't require a nonexistent POTCAR or so that the testing VASP_PSP_DIR is used for those specific tests regardless of whether the user has set up a VASP_PSP_DIR or not.

montoyjh:
will submit PR shortly.

PR #166, opened 2017-10-06
Modify potcar

montoyjh:
Changes the potcar substition to one of the other POTCARs in the test_files that should also exist in a user's VASP_PSP_DIR, if set.  Should address #165.


computron:
Thanks!

PR #168, opened 2017-10-11
Proposal: change schema version to match atomate version

mkhorton:
## Summary

Atomate drone has seen a few changes recently, and version number isn't always changed when the schema changes. Makes sense to me just to track the schema with the atomate version instead?


computron:
Thanks!

PR #169, opened 2017-10-12
Fix percent_delta_volume in vasp.drones.VaspDrone.set_analysis to be a percent

bocklund:
…a percent

## Summary

<Short few sentences, and summary of the major changes in bullet 
points>

* Feature 1
* Feature 2
* Fix 1
* Fix 2

## TODO (if any)

<If this is a work-in-progress, write something about what else needs 
to be done>

* Feature 1 supports a, but not b.

### DIFF
--- a/atomate/vasp/drones.py
+++ b/atomate/vasp/drones.py
@@ -419,7 +419,7 @@ class VaspDrone(AbstractDrone):
         initial_vol = d["input"]["structure"]["lattice"]["volume"]
         final_vol = d["output"]["structure"]["lattice"]["volume"]
         delta_vol = final_vol - initial_vol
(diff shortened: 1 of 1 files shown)


computron:
Thanks, this is certainly correct but there is the issue that if I merge, people will have a database with mixed conventions and no great way to know which convention was used (yes, there is the drone schema, but no one is going to pay attention to that).

What I will do is the following:
* merge your change
* change the name of the fixed key to be "delta_volume_as_percent" - this will be a new key that doesn't interfere with the old convention and people won't have the same key with two different calculation methods mixed in
* update the FixTasksBuilder to update old tasks that use the old convention

PR #171, opened 2017-10-23
Powerup to Preserve Fworker  and include DriftErrorHandler

shyamd:
Changes

1. Little powerup that sets _preserve_fworker
Useful for those of us who spend a few hours writing code to do just this and then realize the 
functionality was already there......
2. Included DriftErrorHandler when ediffg is set.

### DIFF
--- a/atomate/vasp/firetasks/run_calc.py
+++ b/atomate/vasp/firetasks/run_calc.py
@@ -22,7 +22,7 @@ from custodian import Custodian
 from custodian.vasp.handlers import VaspErrorHandler, AliasingErrorHandler, \
     MeshSymmetryErrorHandler, UnconvergedErrorHandler, MaxForceErrorHandler, PotimErrorHandler, \
     FrozenJobErrorHandler, NonConvergingErrorHandler, PositiveEnergyErrorHandler, \
-    WalltimeHandler, StdErrHandler
+    WalltimeHandler, StdErrHandler, DriftErrorHandler
 from custodian.vasp.jobs import VaspJob, VaspNEBJob
 from custodian.vasp.validators import VasprunXMLValidator, VaspFilesValidator
 
@@ -172,6 +172,8 @@ class RunVaspCustodian(FiretaskBase):
         # construct handlers
         handlers = handler_groups[self.get("handler_group", "default")]
 
+        if self.get("ediffg"):
+            handlers.append(DriftErrorHandler())
         if self.get("max_force_threshold"):
             handlers.append(MaxForceErrorHandler(max_force_threshold=self["max_force_threshold"]))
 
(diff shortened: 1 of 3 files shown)


computron:
Hi @shyamd 

Some comments

1. Tests fail (maybe just the requirements need updating in setup.py and requirements.txt?)
2. I would suggest that the DriftHandler within custodian (not atomate) read the INCAR file and try to detect whether it should be activated or not. I am not sure that checking EDIFFG is the best option or not.
3. One of the concepts for atomate is "least surprise". Inserting a DriftHandler whenever there is ediffg, regardless of what the user has set as their handler group, violates that. The DriftHandler is just snuck in under the radar. It is especially surprising since the handler won't be there if the user accepts the default ediffg (None) for RunVaspCustodian, but if they explicitly set that same default value for ediffg as a number rather than None, the handler will magically get added and you can get different otucomes! Instead, you should add the DriftHandler to all the various handler groups - plain for everyone to see. Then, you can filter out the DriftHandler if the computation doesn't look like it's a relaxation type run (some condition, although the ediffg one doesn't look right to me - more like if the job_type is not one of the relaxation types).

shyamd:
1.) Yes the drift correct needs a new version of custodian to be released first, and then update requirements for it. 

Your remaining two points seems to negate each other. Enabling the DriftCorrector as part of handler set but putting the logic for enabling it into custodian would obfuscate the logic to all but the most experienced users who know atomate/custodian intimately.

I vote for adding a flag to be able to not include it, but by default to include it whenever EDIFFG < 0, since that is the only time it is useful.

computron:
I don't want independent flags for everyone who wants a handler. If you want a handler, put it in the list. That is where all the handlers go.

>> Your remaining two points seems to negate each other.

They don't, and here's why. The biggest concern is "do no unintended harm" to the user's code. There are already cases where you can add a custodian handler and it will do nothing. For example, if you set an AliasingErrorHandler but you also set NELECT, the AliasingErrorHandler basically just deactivates. The MeshSymmetryHandler also deactivates itself if ISYM=F or if auto_kpoint generation is used. This logic is there directly in the handler itself to prevent them from doing anything unusual, regardless of whether someone adds the handler through atomate or not.

If the DriftErrorHandler is really only valid/useful for EDIFFG < 0, it can have a similar logic to those other handlers. Then, if someone (not even using atomate) reuses the same handler set as their colleague but switches EDIFFG > 0, your handler shuts itself down gracefully.

No one is confused when a handler doesn't do anything, they are confused when they are unaware that VASP parameters are going to change from what they thought (e.g., set a null handler set to run a "raw" VASP run with custodian double relaxation feature, but still are getting error corrections from some handler that got added somewhere else in the code).

So the plan would be:
- DriftErrorHandler turns itself off if its added in teh context of an invalid parameter set, like other handlers that do the same. This works independent of atomate.
- DriftErrorHandler gets added to the handler_sets where it makes sense in atomate.
- (optional) you can add some code in RunVaspCustodian that removes DriftErrorHandler from the handler_set if it's present there and ediffg >= 0 if you want. This might make things a little more clear for someone reading only atomate code and also will prevent a little computation time from the DriftHandler checking to make sure it's valid when we are pretty confident it won't be. But again, the major issue is handlers sneaking in that you are unaware of.

computron:
Thanks! Will merge as soon as custodian is updated and tests pass

computron:
Thanks!

PR #172, opened 2017-10-26
Added CreateFolder and GrabFilesFromCalcLoc to common.firetasks.glue_… (pull request 1/5)

blondegeek:
This pull requests is in preparation for merging the ferroelectric workflow.

This specific pull requests adds the CreateFolder and GrabFilesFromCalcLoc classes and tests for these classes. Both of these classes will be used by the ferroelectric workflow.

GrabFilesFromCalcLoc may overlap with CopyFiles -- perhaps they should be merged?

### DIFF
--- a/atomate/common/firetasks/glue_tasks.py
+++ b/atomate/common/firetasks/glue_tasks.py
@@ -141,17 +141,24 @@ class CreateFolder(FiretaskBase):
 
     Required params:
         folder_name (str): folder name.
-        change_to (bool): change to new folder.
 
+    Optional params:
+        change_to (bool): change to new folder. Defaults to False.
+        local (bool): whether folder name is relative or absolute. Defaults to True.
     """
-    required_params = ["folder_name","change_to"]
+    required_params = ["folder_name"]
+    optional_params = ["change_to", "local"]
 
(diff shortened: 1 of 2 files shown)


computron:
Hi @blondegeek 

* I think it's fine that CopyFiles is separate from the calc loc  version. This will keep the former one simple.
* Seems strange that you'd need a dedicated Firetask for CreateFolder. But assuming this is really needed the code should look like this

```
new_dir = os.path.join(os.getcwd(), self["folder_name"])
if not os.path.exists(new_dir):
    os.makedirs(new_dir)
if self.get("change_to"):
    os.chdir(new_dir)
```

Comments
- don't surprise the user by creating a folder named test if they don't specify a folder name. The ode should throw an error, that should be a required parameter.
- use os.path.join - avoid forward/back slash incompatibility (Win vs Nix)
- don't need second arg to self.get
- note also that docs should reflect that change_to is an optional param

computron:
Also the CreateFolder should also probably allow for relative vs absolute dirs

computron:
Thanks! I merged but made many changes that will require some updating of any dependent code.

PR #173, opened 2017-10-26
Added GetInterpolatedPOSCAR + test and modified CopyVaspOutputs. (pull request 2/5)

blondegeek:
GetInterpolatedPOSCAR get's the CONTCAR from a two previously run FireWorks and writes a specified interpolated structure as POSCAR.

The modification made to CopyVaspOutputs prevents moving CONTCAR->POSCAR if CONTCAR is included in additional_files and "contcar_to_poscar"=False.

### DIFF
--- a/atomate/vasp/firetasks/glue_tasks.py
+++ b/atomate/vasp/firetasks/glue_tasks.py
@@ -209,25 +209,35 @@ class GetInterpolatedPOSCAR(FiretaskBase):
     Grabs CONTCARS from two previous calculations to create interpolated
     structure.
 
+    The code gets the CONTCAR locations using get_calc_loc of two calculations
+    indicated by the start and end params, creates a folder named "interpolate"
+    in the current FireWork directory, and copies the two CONTCARs to this folder.
+    The two CONTCARs are then used to create nimages interpolated structures using
+    pymatgen.core.structure.Structure.interpolate. Finally, the structure indicated
+    by this_image is written as a POSCAR file.
+
     Required params:
         start (str): name of fw for start of interpolation.
         end (str): name of fw for end of interpolation.
         this_image (int): which interpolation this is.
         nimages (int) : number of interpolations.
-        autosort_tol (float): a distance tolerance in angstrom in which
-          to automatically sort end_structure to match to the closest
-          points in this particular structure.
+
+    Optional params:
+        autosort_tol (float): parameter used by Structure.interpolate.
+          a distance tolerance in angstrom in which to automatically
+          sort end_structure to match to the closest
+          points in this particular structure. Default is 0.0.
+
     """
-    required_params = ["start","end","this_image","nimages","autosort_tol"]
+    required_params = ["start","end","this_image","nimages"]
+    optional_params = ["autosort_tol"]
 
     def run_task(self, fw_spec):
 
         # make folder for poscar interpolation start and end structure files.
-        interpolate_folder = '/interpolate'
-        if not os.path.exists(os.getcwd()+interpolate_folder):
-            os.makedirs(os.getcwd()+interpolate_folder)
-
-            print (os.getcwd()+interpolate_folder)
+        interpolate_folder = 'interpolate'
+        if not os.path.exists(os.path.join(os.getcwd(),interpolate_folder)):
+            os.makedirs(os.path.join(os.getcwd(),interpolate_folder))
 
         # use method of GrabFilesFromCalcLoc to grab files from previous locations.
         CopyFilesFromCalcLoc(calc_dir=None, calc_loc=self.get("start","default"), filenames=["CONTCAR"],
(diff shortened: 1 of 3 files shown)


computron:
I am reopening this PR now that level 1 is merged

computron:
It says there are no conflicts with the base branch, but I don't see my new changes from atomate master (e.g. GrabFilesFromCalcLoc -> CopyFilesFromCalcLoc) reflected here ...

blondegeek:
Just pulled from atomate master.

blondegeek:
Currently fixing.

blondegeek:
Ok! Should be good to go.

atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -202,6 +203,52 @@ def run_task(self, fw_spec):
...
+          to automatically sort end_structure to match to the closest
+          points in this particular structure.
+    """
+    required_params = ["start","end","this_image","nimages","autosort_tol"]
+
+    def run_task(self, fw_spec):
+
+        # make folder for poscar interpolation start and end structure files.
+        interpolate_folder = '/interpolate'
+        if not os.path.exists(os.getcwd()+interpolate_folder):
```
computron: use os.path.join(), not string concatenation. You also don't need slashes in that case, e.g. don't need "/interpolate" just "interpolate"

atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -202,6 +203,52 @@ def run_task(self, fw_spec):
...
+            os.makedirs(os.getcwd()+interpolate_folder)
+
+            print (os.getcwd()+interpolate_folder)
```
computron: don't add print statements; you can add log statements if you want, see logger in other places like ToDbTask

atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -202,6 +203,52 @@ def run_task(self, fw_spec):
...
+
+        # use method of GrabFilesFromCalcLoc to grab files from previous locations.
+        CopyFilesFromCalcLoc(calc_dir=None, calc_loc=self.get("start","default"), filenames=["CONTCAR"],
+                             name_prepend="interpolate/", name_append="_0").run_task(fw_spec=fw_spec)
+        CopyFilesFromCalcLoc(calc_dir=None, calc_loc=self.get("end","default"), filenames=["CONTCAR"],
+                             name_prepend="interpolate/", name_append="_1").run_task(fw_spec=fw_spec)
+
+        # assuming first calc_dir is polar structure for ferroelectric search
+
+        s1 = Structure.from_file("interpolate/CONTCAR_0")
```
computron: you defined a variable earlier called ``interpolate_folder``, use that instead of the hard-coded "interpolate"

atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -202,6 +203,52 @@ def run_task(self, fw_spec):
         return FWAction(stored_data=stored_data)
 
 
+@explicit_serialize
+class GetInterpolatedPOSCAR(FiretaskBase):
+    """
+    Grabs CONTCARS from two previous calculations to create interpolated
```
computron: docs should be clearer about what the code does. The code is creating folders, writing POSCARs, copying CONTCARs, etc. You should write out clearly what the task is going to do.

atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -202,6 +203,52 @@ def run_task(self, fw_spec):
...
+    structure.
+
+    Required params:
```
computron: Many of these parameters don't seem to be required, i.e., the code will run perfectly well without them being set. It's better if you can decide what should really be required and what should really be optional. If it is a required parameter, i.e., you really think the user should be setting it and the code shouldn't be setting an intelligent default, then the code should FAIL if it is not set.

atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -202,6 +203,52 @@ def run_task(self, fw_spec):
...
+        start (str): name of fw for start of interpolation.
+        end (str): name of fw for end of interpolation.
+        this_image (int): which interpolation this is.
+        nimages (int) : number of interpolations.
+        autosort_tol (float): a distance tolerance in angstrom in which
```
computron: I don't understand the ``autosort_tol`` parameter

computron: (although I guess this is a pymatgen pass-through var, so maybe better if it's documented there)

atomate/vasp/firetasks/glue_tasks.py
```diff
@@ -202,6 +203,52 @@ def run_task(self, fw_spec):
...
+        # assuming first calc_dir is polar structure for ferroelectric search
+
+        s1 = Structure.from_file("interpolate/CONTCAR_0")
+        s2 = Structure.from_file("interpolate/CONTCAR_1")
+
+        structs = s1.interpolate(s2,self.get('nimages',5),interpolate_lattices=True,
+                                 autosort_tol=self.get("autosort_tol",0.5))
+
+        # save only the interpolation needed for this run
+        i = self.get("this_image",0)
```
computron: I don't understand why this Task doesn't just write out all 5 (or ``nimages``) interpolations. Why only one of the interpolations?

atomate/vasp/firetasks/glue_tasks.py
computron: (this comment is throughout the code)

atomate/vasp/firetasks/glue_tasks.py
blondegeek: Each interpolation requires both a static and a polarization calculation. For scheduling reasons, I want each interpolation calculation to be performed by a separate FireWork.

atomate/vasp/firetasks/glue_tasks.py
blondegeek: Done.

atomate/vasp/firetasks/glue_tasks.py
blondegeek: I've altered the code so now many of these parameters are required.

atomate/vasp/firetasks/glue_tasks.py
blondegeek: I've added a comment that it is a parameter for Structure.interpolate. I've also set a default and made it an optional parameter.

atomate/vasp/firetasks/glue_tasks.py
blondegeek: Done.

atomate/vasp/firetasks/glue_tasks.py
blondegeek: Done.

atomate/vasp/firetasks/glue_tasks.py
blondegeek: Done.

computron:
Thanks! Merging

I'll make some minor changes myself after merge - mainly, if a parameter is required, don't use self.get("required_param") since that will create a default value of None if unset, which won't cause the code to immediately fail and the failure will occur somewhere else in the code. Just use self["required_parameter"] so the code will immediately fail right then and there if the required parameter is not set.

Basically, the job of the code is not to avoid failures, but rather to ensure the code clearly tells the user where his/her mistakes are (not have the code fail 10 lines downstream of their mistake).

PR #179, opened 2017-11-13
Parse LOCPOT

shyamd:
Simple addition to the VASP drone to parse LOCPOT and save the axis averages which can be used for surface and defect corrections. 

Currently defaults to not parse, but can very easily change to parse by default since it's a tiny amount of data that might be useful for other things?

### DIFF
--- a/atomate/vasp/tests/test_drones.py
+++ b/atomate/vasp/tests/test_drones.py
@@ -12,6 +12,8 @@ from pymatgen.io.vasp import Outcar, Oszicar
 
 from atomate.vasp.drones import VaspDrone
 
+import numpy as np
+
 
 module_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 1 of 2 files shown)


atomate/vasp/tests/test_drones.py
```diff
@@ -126,5 +128,19 @@ def test_detect_output_file_paths(self):
                               'wavecar': 'WAVECAR.relax1.gz'},
                              doc['calcs_reversed'][1]['output_file_paths'])
 
+    def test_parse_locpot(self):
+        drone = VaspDrone(parse_locpot=True)
+        doc = drone.assimilate(self.Si_static)
+
+        self.assertTrue(drone.parse_locpot)
+        self.assertTrue('locpot' in doc['calcs_reversed'][0]['output'])
+        self.assertTrue(0 in doc['calcs_reversed'][0]['output']['locpot'])
```
computron: are these 3 checks needed? seems like the tests afterwards which reference these indices will automatically check for this?

computron: (referring to checks of form ``self.assertTrue(0 in doc['calcs_reversed'][0]['output']['locpot'])``

computron:
For parse by default, can you give me an idea of:

* how much data is it? (is it just the 3 numbers or more than that)
* does it / can it take a long time to parse?

atomate/vasp/tests/test_drones.py
shyamd: You're right that the next three tests cover any issues with the indicies.  These three don't hurt and provide more explicit checks for what is wrong.

shyamd:
The LOCPOT grid is based on NGF, and from a cursory review of calcs i've run that ranges from 30-40 to as much as 100 numbers per axial direction, as much as 3-4 kb of floats added to the document.

computron:
Ok I'm just going to merge as-is for now. If you feel strongly that the default should be True, you should submit another PR and I'll likely just merge w/o further discussion.

PR #180, opened 2017-11-13
Parent structure

montoyjh:
## Summary

A few minor changes in preparation for some larger ones involving builders.

* Added optional kwarg "parent_structure" to TransmuterFW that adds parent_structure and metadata to "additional_fields" for task document.  Defaults to false but set to true for deformations tasks.
* Refactored structure metadata generation to get all data required by builder recognition of "parent structure"
* Added support for Fourth order elastic constants (i.e. a sufficient set of strain states and removed error raised)
* Changed "fw_spec_field" in VaspToDb to "fw_spec_fields" to allow for more than one field to be included.


atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -229,7 +219,7 @@ class ElasticTensorToDb(FiretaskBase):
         db_file (str): path to file containing the database credentials.
             Supports env_chk. Default: write data to JSON file.
         order (int): order of fit to perform
-        fw_spec_field (str): if set, will update the task doc with the contents
+        fw_spec_fields (str): if set, will update the task doc with the contents
```
computron: note - this should be [str] now

atomate/vasp/fireworks/core.py
```diff
@@ -445,10 +446,9 @@ def __init__(self, structure, magmom, name="spin-orbit coupling",
 
 class TransmuterFW(Firework):
```
computron: I'm confused by these changes to TransmuterFW. In particular:

* the parent_structure argument is not documented, so I'm not sure what this is. The name of the parameter would make it seem like a Structure object, but it seems to be a bool that adds metadata. Why is not called ``add_metadata`` or something?
* I think I'm on board with the removal of the ``parents`` parameter. From what I can tell, ``**kwargs`` should handle that already. Is that correct? If so, I would suggest not changing it in this PR but submitting a separate PR that just cleans up all the extraneous ``parents`` parameters in all the various FWS (there are many that contain this)

computron:
See my (minor) comments, let me know when finished making changes

montoyjh:
Sorry this took so long, I believe I've addressed all issues, let me know if any further.

montoyjh:
I'm having some issues with this in practice (mostly related to any transmuter fireworks that are derived from prior fireworks), so I'm going to close and submit a new PR when I have a better idea of how best to implement this.

PR #181, opened 2017-11-14
Initial versions of magnetic workflows

mkhorton:
## Summary

Some workflows for magnetic orderings and magnetic deformation.

These were actually written some time ago, but I was waiting for the necessary pymatgen code to be merged + become stable first (there were some changes to MagOrderingTransformation, etc.)

For anyone wanting to run magnetic orderings, if it's a simple material and/or you know what elements are likely to be magnetic, and you have a good idea how it works, etc., I would just use a simple TransmuterFW and the MagOrderingTransformation directly. This workflow is more for when you don't know much about the material to start with.

The workflow will output a summary to a dedicated collection in a db. In future, there's no intrinsic reason this functionality can't be wrapped up into the materials builder (i.e. treat different magnetic orderings like any other task doc), but for stand alone use, outputting the summary to a dedicated collection seems to work well.

## TODO (if any)

* I'll be adding some unit tests + test files soon -- by all means wait to merge the PR if you'd like to wait for the unit tests first. Workflows have been tested though and seem to work.

* More testing for ferrimagnetic orderings needs to be done... this is fairly new.

* Some other miscellaneous things, expect some future PRs for small fixes/changes


mkhorton:
(Test failures are due to a yaml file that's missing from pymatgen's pip distribution, even though it's in the main repo... this is causing the CI tests to fail, even though they'd work if pymatgen was installed manually. Will try and find out why this is happening!)

computron:
Ok, I never really understood Python packaging. But to get non-Python files to appear I usually make sure they are there in setup.py's ```package_data``` parameter. That parameter is a bit finicky and you need to format it correctly.

Will review this PR once the tests pass ...

shyuep:
What is the specific yaml file that was missing?

mkhorton:
It's `pymatgen/analysis/magnetism/default_magmoms.yaml` -- however, since I made this, I noticed you've put magmoms in their own `VASPIncarBase.yaml`, so I was thinking about deleting this anyway and just using the VASPIncarBase instead.

mkhorton:
Closing for now until I've added some test files + a few minor changes

PR #182, opened 2017-11-22
Initial version of ATAT / SQS workflow

mkhorton:
## Summary

Minimal working version of SQS workflow.

Note that this is slightly unusual because the `mcsqs` sometimes will never terminate (it's a Monte Carlo code, will just keep trying permutations indefinitely). As such, the FW has a `walltime` kwarg to force termination after a specified amount of time. Not sure if there's a more elegant way of doing this.

## TODO (if any)

* Transformation should be added to pymatgen to re-combine SQS template with original disordered structure for appropriate provenance ... still need to code this

* Some functions need unit tests, not sure the best way to do this since they rely on a database being present -- mongomock?


bocklund:
Hey @mkhorton, I have a little bit of code [here](https://github.com/PhasesResearchLab/prlworkflows/blob/master/prlworkflows/sqs_db.py) and [here](https://github.com/PhasesResearchLab/prlworkflows/blob/master/prlworkflows/sqs.py) that deals with parsing the ATAT format for SQS (the first link) and creating some abstract SQS object that subclasses pymatgen structures and allows for enumeration of all the possible concrete structures from the abstract one.

I'm not sure if any of that would be helpful, but feel free to ping me if you want me to incorporate this at all. The code in that repo is more built around using SQS with atomate once they are generated, rather than generating the SQS themselves. However, I am very interested and thankful for what you are doing here. Let me know if there's anything I can do to help.

mkhorton:
Hi @bocklund, many thanks for the links they look very useful! Do you have any plans to add any of that code to pymatgen? There's now a `pymatgen.io.atat` module, which is just very basic I/O currently. The plan was to add an SQSTransformation to `pymatgen.transformations.advanced_transformations` next, purely for provenance (so we can treat it like any other Transformation).

The intent with this atomate workflow was to make it easier to make our own SQS db, so hopefully we could on-demand generate SQS templates for some of the less common structures.

I'd be happy to incorporate any comments/suggestions you may have for this workflow. One issue is that we currently include the spectator atoms in the SQS generation (under the logic that these will change what clusters are generated, e.g. for ABO3 perovskites we include the O3), but it seems from your code that this might not be necessary, and just considering the disordered sublattice is sufficient? Are there any best practices you would suggest we follow? Happy to chat over email about this too.

mkhorton:
@computron this should be ready to merge now, unless there are any comments/changes requested -- it's pretty minimal but is tested + working.

Future additions will require additional SQS support in pymatgen.

mkhorton:
Thanks for the review @montoyjh!

atomate/atat/database.py
```diff
@@ -0,0 +1,33 @@
...
+# based on FeffCalcDb
+
+logger = get_logger(__name__)
+
+class McsqsCalcDb(CalcDb):
+
+    def __init__(self, host="localhost", port=27017,
+                 database="sqs", collection="tasks",
+                 user=None, password=None):
+        super(McsqsCalcDb, self).__init__(host, port, database, collection, user, password)
```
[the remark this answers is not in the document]
computron: We should decide on whether the various names should look like McsqsCalcDb, McsqsDrone, etc. or like SQSCalcDb, SQSDrone, etc.

Happy to defer to your judgement, just think about it

atomate/atat/database.py
```diff
@@ -0,0 +1,33 @@
...
+from atomate.utils.utils import get_logger
+
+__author__ = 'Matthew Horton'
+__credits__ = 'Anubhav Jain, Kiran Mathew, Josua Vieten'
+__email__ = 'mkhorton@lbl.gov'
+# based on FeffCalcDb
+
+logger = get_logger(__name__)
+
+class McsqsCalcDb(CalcDb):
```
computron: We should decide on whether the various names should look like McsqsCalcDb, McsqsDrone, etc. or like SQSCalcDb, SQSDrone, etc.

Happy to defer to your judgement, just think about it

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+        structures at random); however, whatever structure
+        it finds in this time (the 'bestsqs') is often
+        still useful.
+
+        :param disordered_struct: disordered structure
+        :param name: name of the Firework
+        :param size: ratio of atoms in the output supercell
+        to number of atoms in input cell, if None will
+        choose the smallest possible supercell
+        :param clusters: dict of cluster sizes, e.g.
```
computron: I am a bit afraid that the clusters parameters (if set manually and rather than using the default of None) will not serialize and deserialize properly. In particular, Mongo doesn't like int keys for dicts and we end up having to store such keys as strings and convert back to int() in the from_dict(). Would check this ...

computron: (this is limitation is in the JSON spec?)

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+        pick sensible defaults, the value of mcsqs is
+        that this should converge rapidly with number of
+        clusters (aim for around a dozen)
+        :param user_input_settings: dict of additional keyword
+        value pairs to pass to mcsqs, such as Monte Carlo
+        temperature, no validation is performed on this dict
+        :param: max_denominator (int):
+        Will round occupancies to the nearest rational number,
+        with the maximum denominator as specified. Use None not
+        to do perform rounding, though not mcsqs might not like this!
```
computron: typo ... "to do perform"

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+        :param: walltime (int): Time to run mcsqs for in minutes,
```
computron: can you rename to walltime_mins?

A lot of our "time" parameters are in seconds. This will just clarify bit

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+        :param: ncores (int): number of instances of mcsqs
+        to run (mcsqs is not parallelized, these are run
+        independently of each other), by default will try
+        to detect automatically
+        :param: db_file:
+        :param kwargs: Other kwargs that are passed to Firework.__init__.
+        """
+
+        if walltime < 3:
+            raise ValueError("Please increase walltime.")
```
computron: more helpful message

"Walltime must be at least 3"
or
"Walltime must be at least 3, got {}".format(walltime_mins)

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if disordered_struct.is_ordered:
+            raise ValueError("You must input a disordered structure.")
```
computron: more helpful message

"The structure provided looks to be ordered; please input a disordered structure."

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if db_file is None:
+            logger.warn("Database not specified for McsqsFW.")
```
computron: why a warning? I thought this was a valid procedure

mkhorton:
Thanks very much for the review @computron -- I'll be incorporating your comments and Joey's comments + will update the PR.

It's also come to my attention that the output needs some additional error checking for detecting amorphous-like (i.e. failed) outputs. This is better suited to Custodian I think. Will have to investigate further on how best to do this.

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+        generate_cluster_cmd = "mcsqs {}".format(cluster_str)
+
+        if user_input_settings:
+            user_input_settings_str = " ".join(["{}={}".format(k, v) for k, v
+                                            in user_input_settings.items()])
+        else:
+            user_input_settings_str = ""
+
+        # using same approach as Custodian to detect ncores
+        ncores = os.environ.get('NSLOTS') or multiprocessing.cpu_count()
```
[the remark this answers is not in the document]
computron: I think this is fine, I doubt anyone is going to want to tune this. I am more concerned that it doesn't seem the ncores parameter from the __init__() is respected? Might just be missing it.

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+        temperature, no validation is performed on this dict
+        :param: max_denominator (int):
+        Will round occupancies to the nearest rational number,
+        with the maximum denominator as specified. Use None not
+        to do perform rounding, though not mcsqs might not like this!
+        :param: walltime (int): Time to run mcsqs for in minutes,
+        1 minute will be deducted from this to run the database
+        tasks.
+        :param: ncores (int): number of instances of mcsqs
+        to run (mcsqs is not parallelized, these are run
```
computron: Use Google-style docstrings

mkhorton: (Note for future ref, I just found `pyment` is good for auto-converting docstrings -- I converted the propnet ones with it. The default PyCharm style (this style) is based on [PEP 287](https://www.python.org/dev/peps/pep-0287/) though, so should be picked up by any documentation generators fine, but agree it should be changed for consistency.)

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+        if not timeout_cmd:
+            raise RuntimeError("Requires timeout to limit walltime.")
+
+        with open(os.path.join(module_dir, 'run_mcsqs_template.sh')) as f:
+            template = f.read()
+
+        bash_script = Template(template)
+        bash_script = bash_script.substitute({
+            'timeout_cmd': timeout_cmd,
+            'time': 60*(walltime - 1),  # seconds, minus 60 seconds for DB task
```
computron: I think you should rephrase the walltime parameter to only be the SQS time. In the docs for walltime just say that one should leave at least 1 minute for the db insertion.

atomate/atat/fireworks/core.py
bocklund: FYI you can set the PyCharm docstring style per project at Settings > Tools > Python Integrated Tools to Numpy, rst, Google, etc. https://www.jetbrains.com/help/pycharm/python-integrated-tools.html

atomate/atat/fireworks/core.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @staticmethod
+    def _determine_min_cell(disordered_struct):
+        """
+        Get a minimum cell size for an ordered structure.
+        """
+        denoms = {}
+        for sp_occu in disordered_struct.species_and_occu:
+            for sp, occu in sp_occu.items():
+                denom = Fraction(occu).limit_denominator(100).denominator
```
computron: not sure why this is hard-coded as 100 rather than using the max_denominator set earlier in the FW

atomate/atat/utilities.py
```diff
@@ -0,0 +1,128 @@
...
+from pymatgen.analysis.structure_matcher import StructureMatcher, OccupancyComparator
+
+from atomate.atat.database import McsqsCalcDb
+
+__author__ = 'Matthew Horton'
+__credits__ = 'Josua Vieten'
+__email__ = 'mkhorton@lbl.gov'
+
+
+def preprocess_disordered_structure(disordered_struct, max_denominator=8):
```
computron: Should some of this go in pymatgen?

atomate/atat/utilities.py
```diff
@@ -0,0 +1,128 @@
...
+    disordered_struct.scale_lattice(1.0)
+
+    if max_denominator:
+        trans = DiscretizeOccupanciesTransformation(max_denominator)
+        disordered_struct = trans.apply_transformation(disordered_struct)
+
+    return disordered_struct
+
+
+def retrieve_bestsqs_template(disordered_struct, db_file,
```
computron: should this go in McsqsCalcDB instead of utilities?

computron:
I also made some comments

Please let me know when they are addressed and I will check this again

atomate/atat/fireworks/core.py
mkhorton: Yeah I'm not sure why either. This function should really go in pymatgen somewhere anyway (e.g. as a method on disordered structures generally).

computron:
(btw, more elegant than walltime would be some kind of convergence criterion I guess)

mkhorton:
Yeah, agreed -- this is something else well-suited for Custodian. Really need to get more practical examples running to get a good sense what convergence criterion to use though.

atomate/atat/fireworks/core.py
shyuep: Isn't this already the case? We already coded a proper disordered occupancy rounding in pymatgen.

mkhorton: Yeah, there's DiscretizeOccupancyTransformation there now, but I meant the method for determining the smallest ordered cell size for a disordered structure -- I don't think that's in pymatgen?

computron:
@mkhorton can you give a short update on this PR? I usually start closing PRs and ask people to resubmit later after a month of inactivity to keep things tidy

mkhorton:
Sure, by all means close for now.

Things that need to happen:

* Tidy up of code in response to code review (nothing major just haven’t
had chance yet)

* A check for amorphization (to see if the SQS has failed)

* Maybe some smarter defaults (pending more real world testing)

This isn’t for a current project, so it’s waiting for some free time.

computron:
It's been about 2 months and this PR is pretty out of date, so I am closing. Please request to re-open or submit a separate PR when the code is ready.

PR #183, opened 2017-11-28
force gamma in DFPTFW added

dbroberg:
## Summary
DFPTFW requires force_gamma to be set to True. If one does not copy from a previous calculation and then DFPTFW previously broke from lack of forced gamma pt calculation.

* force_gamma feature for DFPTFW added

## TODO (if any)

None.

### DIFF
--- a/atomate/vasp/fireworks/core.py
+++ b/atomate/vasp/fireworks/core.py
@@ -317,7 +317,7 @@ class DFPTFW(Firework):
         if copy_vasp_outputs:
             t.append(CopyVaspOutputs(calc_loc=True, contcar_to_poscar=True))
             t.append(WriteVaspStaticFromPrev(lepsilon=True, other_params={
-                'user_incar_settings': user_incar_settings}))
+                'user_incar_settings': user_incar_settings, force_gamma=True}))
         else:
             vasp_input_set = MPStaticSet(structure, lepsilon=True, force_gamma=True,
                                          user_incar_settings=user_incar_settings)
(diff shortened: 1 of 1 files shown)


computron:
Thanks! Before I merge, can you explain why force_gamma is needed only when there is no previous calc? i.e., why does the fact that there is a previous calc affect whether you need to turn gamma point on or not?

dbroberg:
Yes - gamma point is always required for an LEPSILON=True calculation because of "static" electric fields being applied (long wavelength-type arguments). 

The reason this bug has not been caught previously is likely because an OptimizeFW job preceded it and was used to copy the previous calc over to DFPTFW. The default setting for OptimizeFW is force_gamma=True, so this never broke anything.

computron:
Ok - although it's possible that previous calculations will have not set gamma=T, e.g.:

* someone manually turned off force_gamma in the previous calc
* the previous calc was not run through atomate
* a custodian rule changed gamma to monkhorst mid-run

To account for this, can you also change the line above your change to include "force_gamma" to look like this?
```
t.append(WriteVaspStaticFromPrev(lepsilon=True, other_params={ 'user_incar_settings': user_incar_settings, 'force_gamma': True}))
```

Then even the WriteVaspStaticFromPrev should force gamma. After that I'll merge

computron:
pretty sure force_gamma should have quotes around it?

computron:
Also equals should be colon ... just look at the code I wrote above?

dbroberg:
Sorry about that - should be better now.

computron:
great, thanks! merging

PR #188, opened 2017-12-08
add more control to BoltztrapFW to directly take doping, tgrid and tm…

albalu:
add more control to BoltztrapFW to directly take doping, tgrid and tmax as args. The default values are set so that the change is backwards-compatible and a real Si boltztrap workflow has been tested.


computron:
Thanks!

PR #189, opened 2017-12-11
Update to latest pymatgen and SCAN structure optimization

shyamd:
Edit:
Had to update to latest pymatgen to get the new MVLScanRelaxSet. As a result had to fix some other issues to make everything work

- Moved get_uri and get_database from pymatgen-db which has other conflicts with pymatgen itself. These two are simply utility functions and atomate doesn't depend on pymatgen-db at all anymore.
- fixed a bug in the VaspToPolarizationDB task where sort was being called on a zip generator. 
- Added in default SCAN optimization workflow
- Added in firetask to clean up files


computron:
Note - I am waiting for the tests to pass before checking this

shyamd:
Understood. As soon as pymatgen gets a new release it should work.

atomate/utils/utils.py
```diff
@@ -311,4 +312,37 @@ def recursive_update(d, u):
 def get_a_unique_id():
     ts = "{:.4f}".format(time())
     ts += str(randint(0,9999)).zfill(4)
-    return ts
\ No newline at end of file
+    return ts
+
+
+def get_uri(dir_name):
```
computron: * should probably go into utils/fileio.py
* make a static method of FileClient? (is this compatible with FileClient?)

atomate/utils/utils.py
```diff
@@ -311,4 +312,37 @@ def recursive_update(d, u):
...
+    fullpath = os.path.abspath(dir_name)
+    try:
+        hostname = socket.gethostbyaddr(socket.gethostname())[0]
+    except:
+        hostname = socket.gethostname()
+    return "{}:{}".format(hostname, fullpath)
+
+
+
+def get_database(config_file=None, settings=None, admin=False, **kwargs):
```
computron: why move this out of pymatgen-db and into atomate?

atomate/vasp/drones.py
```diff
@@ -103,7 +104,7 @@ def __init__(self, runs=None, parse_dos="auto", compress_dos=False, bandstructur
         self.compress_dos = compress_dos
         self.additional_fields = additional_fields or {}
         self.use_full_uri = use_full_uri
-        self.runs = runs or ["relax" + str(i + 1) for i in range(9)]  # can't auto-detect: path unknown
+        self.runs = runs or ["relax" + str(i) for i in range(9)]  # can't auto-detect: path unknown
```
computron: do we have relax0?

I thought we had relax1, relax2, etc.

atomate/vasp/workflows/presets/core.py
```diff
@@ -53,6 +53,34 @@ def wf_bandstructure(structure, c=None):
 
     return wf
 
+def wf_scan_opt(structure,c={}):
```
computron: does this need to be its own workflow? Can it be a powerup to other workflows?

I don't want to end up with wf_scan_static, wf_scan_elastic_tensor, wf_scan_dielectric, etc. - and then repeat that with every other functional.

atomate/vasp/workflows/presets/core.py
```diff
@@ -53,6 +53,34 @@ def wf_bandstructure(structure, c=None):
...
+                                            "half_kpts_first_relax":half_kpts,
+                                            "job_type":"metagga_relaxation_run",
+                                            "db_file":db_file,
+                                            "vasp_cmd":vasp_cmd})
+    wf = add_common_powerups(wf, c)
+
+    if c.get("ADD_WF_METADATA", ADD_WF_METADATA):
+        wf = add_wf_metadata(wf, structure)
+
+    if not c.get("KEEP_WAVECAR", False):
```
computron: I'd like for workflows to try to keep the same sorts of "defaults" to prevent surprises and inconsistencies when transitioning. e.g., going from PBE to SCAN should just change PBE to SCAN, not start adding in other logic and preferences by default. Also, I prefer that workflows be as "plain" as possible unless the user requests otherwise. Otherwise I end up with a whole lot of help tickets as to why they ran 50 workflows with atomate but realized the WAVECAR doesn't seem to be retained.

Thus you should:
* rename/invert KEEP_WAVECAR to REMOVE_WAVECAR (the latter is more of the active step you are taking)
* default REMOVE_WAVECAR to False, since this is "least surprise"
* allow a config.py option to have REMOVE_WAVECAR always be True in case someone wants to take the active step of removing WAVECARS and don't want to set the parameter each time
* copy-paste the same code into all preset workflows where WAVECAR is likely unimportant

atomate/utils/utils.py
computron: Nevermind I see you answered this already.

Can you remove pymatgen-db from requirements.txt and setup.py in that case?

requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy==1.13.3
 scipy==1.0.0
 FireWorks==1.6.1
-pymatgen==2017.12.8
+pymatgen==2017.12.15
```
computron: you should also update setup.py

computron:
Thanks! There's a lot of changes so I did a code review - let me know when you have gone through and want me to look at it again

atomate/utils/utils.py
shyamd: I don't think it is. FileClient needs the username and server address. While this gets a hostname, often we get the compute node address which is mostly useless. I think on further inspection, I would advocate eliminating it from the drone.

atomate/vasp/drones.py
shyamd: The scan_opt_job that I put into custodian has a relax0 task. This is just a pre-calculation of the Wavefunctions using regular GGA that is used to condition the electronic SCF cycle when running a metaGGA calculation. This was just the simplest solution between the two and @shyuep  didn't complain about it there. I can change both so that this step is named something else:  "e_condition" ?

atomate/vasp/workflows/presets/core.py
shyamd: Noramally, the user sets LWAVE to true and gets a WAVECAR when they want. With the preconditioning step, custodian leaves behind several copies of WAVECARs from the original GGA opt. This is there to clean that up.  I think I may be able to integrate this into custodian now that I think about it.

requirements.txt
shyamd: Done

atomate/vasp/workflows/presets/core.py
shyamd: Right now, only the structure optimization step benefits from the extra logic embedded in custodian. So far, I haven't needed the electronic preconditioning for static calculations. For every other type of calculation, I don't know if it will work or is needed. Do we want to examine the custodian logic? Should there be a default GGA precondition flag or something like that in RunVaspCustodian rather than the current method of having a special custodian job? I'm going off my current experience with SCAN and don't see a point in putting in something into atomate that someone has actually tried and tested.

shyamd:
Made some individual comments and some small changes. Currently there is a metagga_opt_run in custodian that is just the double relaxation with the GGA precondition step, making it a three step calculation. I haven't needed it for static calculations, and haven't tried with other calculations. I'm hesitant to make it a generic powerup and then have someone complain about their SCAN ferroelectric workflow or NEB workflow not converging when we haven't even tried that.

computron:
@shyamd next step is to fix the Codacy issues, most notably the mutable default args

Still thinking about the SCAN optimization workflow. At the very least I'm thinking this should get moved to presets/scan.py in order to create a new "branch" of workflow presets for SCAN

computron:
I think this is almost good to go - see my most recent comments, all pretty minor

computron:
why closed?

shyamd:
OH! I didn't realize this was still around. I accidentally deleted the local and remote branch on my end and saw the PR disappear, so I just made a new one.

PR #190, opened 2017-12-15
optional parameters in CopyFilesFromCalcLoc

albalu:
## Summary

1) two args "name_prepend", "name_append" were coded to be optional but listed as required parameters.
2) added optional arg "keep_filenames" that is necessary if copying from subfolders. I may need this in subsequent AmsetFW PR. keep_filenames defaults to False so it is backwards-compatible. Please feel free to revert this commit if there's an issue with this.

### DIFF
--- a/atomate/common/firetasks/glue_tasks.py
+++ b/atomate/common/firetasks/glue_tasks.py
@@ -89,7 +89,7 @@ class CopyFilesFromCalcLoc(FiretaskBase):
         name_append (str): string to append to destination filenames.
         keep_filenames (bool): if True, all copied filenames are kept the same.
             The difference w/ not setting name_prepend and name_append is that
-            "foo/bar" can be in filenames copy from subfolder "foo" in calc_dir
+            "foo/bar" can be in filenames copied from subfolder foo in calc_dir
     """
 
     required_params = ["calc_loc", "filenames"]
--- a/atomate/common/firetasks/glue_tasks.py
+++ b/atomate/common/firetasks/glue_tasks.py
@@ -87,13 +87,10 @@ class CopyFilesFromCalcLoc(FiretaskBase):
     Optional params:
         name_prepend (str): string to prepend filenames, e.g. can be a directory.
         name_append (str): string to append to destination filenames.
-        keep_filenames (bool): if True, all copied filenames are kept the same.
-            The difference w/ not setting name_prepend and name_append is that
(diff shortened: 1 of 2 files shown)


computron:
As discussed on the phone, you should

- modify $ALL to find all subdirectories too
- add an $ALL_NO_SUBDIRS to find files but not subdirs
- remove the keep_filenames parameter

albalu:
Please note that I kept the default behavior (filenames is None) as it was before which is equivalent to '$ALL_NO_SUBDIRS' in filenames

computron:
There are a couple of things I don't understand:

1. If I select "$ALL", it looks like it just does a direct copy and doesn't respect other parameters like ``name_prepend`` or ``name_append``. If those parameters are set, the code should throw an error and the documentation should make it clear that $ALL is not compatible with those parameters.
2. It is not clear to me how this code will function properly when ``filenames`` is None. Maybe I'm not looking at correctly. Can you explain?

albalu:
Hi @computron 
1) I made sure they are not compatible and made it clear in docs
2) filenames is None is currently redundant; it is the same as '$ALL_NO_SUBDIRS'. I kept it to keep the default behavior the same as before and have it backward-compatible

computron:
Hi Alireza,

Regarding 2 - I know your intent is for None to be the same as '$ALL_NO_SUBDIRS'. What I was trying to say is that, upon reviewing the actual code, I don't see that to be the case. Here is the relevant piece of code:

```
        filenames = self.get('filenames', None)
        if filenames:
            if isinstance(self["filenames"], six.string_types):
                raise ValueError("filenames must be a list!")

        if '$ALL_NO_SUBDIRS' in filenames:
            files_to_copy = fileclient.listdir(calc_dir)
        elif '$ALL' in filenames:
            if self.get('name_prepend', None) or self.get('name_append', None):
                raise ValueError('name_prepend or name_append options not compatible with "$ALL" option')
            fileclient.copytree(calc_dir, os.getcwd())
            return
        else:
            files_to_copy = filenames
```

When I look at the code, it looks like if you set None to the filename, the code will crash when it gets to ``if '$ALL_NO_SUBDIRS' in filenames``.

Can you confirm that you have tested this code with filename as None? If not, can you explain why you think the code should work when filename is None. I just don't see it.

albalu:
You are right, the code was wrong; sorry about that. This new one is good; I just tested all cases.

computron:
Thanks!

PR #191, opened 2017-12-17
apply vasp_input_set_params when StaticFW have parents

specter119:
## Summary

apply vasp_input_set_params when StaticFW have parents
> current StaticFW will not apply vasp_input_set_params when have parents


computron:
Thanks!

PR #197, opened 2018-01-17
Updated FW construction

shyamd:
## Summary
Updated FW construction for the base FWs so that they dont have to have structures anymore. Structure or parent+calc_locs is now enforced via errors on construction.

Minor change
- Updated TransmuterFW to write transformations.json so drone can pick it up

Edit: Don't commit this just yet. I'm going to add in some tests for the new behavior.


computron:
You should use the ``add_additional_fields_to_taskdocs`` powerup for this

computron:
Can you change it so that the format of the name is retained, but the composition shows up as "unknown" (or something similar)

That way there is always a consistent *format* for the fw name regardless of whether the composition is known or not, and we don't have different fw name formats.

computron:
Thanks!

PR #200, opened 2018-01-23
Minor change to drone and change allowing for FOEC

montoyjh:
## Summary

Adds final stress and forces to the output field for drone schema.  Also adds small bit of functionality enabling fourth-order elastic constant calculation.


computron:
Thanks!

Issue #203, opened 2018-02-02
AdsorptionTests fail unit tests


computron:
@montoyjh (?)

The adsorption workflow fails the unit tests. Unfortunately, CircleCI is not properly configured to run VASP simulation tests (e.g., POTCARs required and such) which is why this was not caught earlier but that is a separate issue.

I am pretty sure this is a simple fix(?). The EDIFFG for MVLSlabSet looks to have changed to -0.02. So either the MVLSlab set default or the test needs to be changed. At least that's what a casual look at the code seems like.

======================================================================
FAIL: test_wf (atomate.vasp.workflows.tests.test_adsorbate_workflow.TestAdsorptionWorkflow)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/ajain/Documents/code_matgen/atomate/atomate/vasp/workflows/tests/test_adsorbate_workflow.py", line 68, in test_wf
    assert all([vis.incar['EDIFFG']==-0.01 for vis in ads_vis])
AssertionError: 
-------------------- >> begin captured logging << --------------------
launchpad: DEBUG: RESTARTED fw_id, launch_id to (1, 1)
launchpad: INFO: Performing db tune-up
launchpad: DEBUG: Updating indices...
launchpad: INFO: LaunchPad was RESET.
pymatgen.core.surface: DEBUG: (1, 1, 1) has 1 slabs... 
pymatgen.core.surface: DEBUG: (1, 1, 0) has 1 slabs... 
pymatgen.core.surface: DEBUG: (1, 0, 0) has 1 slabs... 
launchpad: DEBUG: RESTARTED fw_id, launch_id to (1, 1)
launchpad: INFO: Performing db tune-up
launchpad: DEBUG: Updating indices...
launchpad: INFO: LaunchPad was RESET.
--------------------- >> end captured logging << ---------------------

montoyjh:
I'll take a look.

computron:
FYI pretty sure the above error is still there even though the PR was merged

computron:
@ardunn is this test passing after fixing the CircleCI testing? If so, can you close this issue?

computron:
I am pretty sure this works now, closing the issue

PR #207, opened 2018-02-06
added fix for drones with selective dynamics criteria

montoyjh:
## Summary

The issue with the adsorption workflow is actually related to whether or not selective dynamics is included in determining whether the maximum force threshold is reached.  Presumably, if a VASP user has enabled selective dynamics (which appear as a site_property in the final structure of a parsed Vasprun), they don't want those forces which are ignored in the optimization included in the determination of max force, so I just filter the max force by selective dynamics if they're present in the final structure object site_properties.

Should address #203.


computron:
Thanks!

PR #211, opened 2018-02-15
Fix get_database

shyamd:
## Summary

* Fixes get_database so that it works
* Add in tests for get_database and get_uri

### DIFF
--- a/atomate/common/firetasks/tests/test_parse_outputs.py
+++ b/atomate/common/firetasks/tests/test_parse_outputs.py
@@ -36,7 +36,6 @@ class TestDrone(AbstractDrone):
 class TestToDbTask(AtomateTest):
 
     def test_ToDbTask(self):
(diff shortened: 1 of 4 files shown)


computron:
@shyamd see Codacy message

Does UtilsTest need to subclass AtomateTest or can it subclass unittest.TestCase as before? If it really needs to subclass AtomateTest you should make the setUp function match (have an lpad arg)

shyamd:
@computron , I'm making it subclass AtomateTest because I'm pretty sure MongoDB isn't running on the CircleCI image that is currently being used. As a regular unitTest, it fails with an can't connect to MongoDB error message: "pymongo.errors.ServerSelectionTimeoutError: localhost:27017: [Errno 111] Connection refused".  The tests pass on my comp when MongoDB is running and fails in the same way when MongoDB is not. I asked @shyuep to chime in on the issue, since it seems to be one he setup for pymatgen.

PR #214, opened 2018-03-12
Drone Updates

shyamd:
## Summary

Primarily to fix a bug where some basic bandgap properties weren't being copied to the root output document because is_metal wasn't present in the calculation's output. 

- Added some additional comments along the way
- Removed compress_bs and compress_dos, since those didn't actually do anything and the DB is responsible for compressing them.


computron:
Thanks!

PR #216, opened 2018-03-21
wrong dash breaks `python setup.py develop`

tschaume:
```
  File "setup.py", line 15, in <module>
    long_description=open(os.path.join(module_dir, 'README.md')).read(),
  File "/global/u1/m/matcomp/mp_prod_atomate/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 912: ordinal not in range(128)
```


computron:
Thanks!

PR #221, opened 2018-04-19
fix AttributeError: 'bool' object has no attribute 'lower' for VaspToDb

specter119:


computron:
Thanks for pointing this out. You revision will unintentionally raise an error if DEFUSE_UNSUCCESSFUL is False, which was not the intended behavior.

I have fixed the issue in a separate commit so am closing this one.

Thanks!

PR #223, opened 2018-04-23
DB Updates

shyamd:
## Summary

Some updates for  the drone and db

- Removes force checking from the drone
- Stores task_id and compression metadata in GridFS for band structures and DOS


computron:
Thanks!

PR #225, opened 2018-04-27
support list mutiply for ModifyIncar

specter119:
support list mutiply for ModifyIncar


computron:
Thanks! 

Can you just modify this line:
```
if isinstance(incar[k], list):
```

to read:
```
if hasattr(incar[k], '__iter__'):  # is list-like
```

The latter is considered more Pythonic and it will check for all kinds of iterable data structures like list or tuple in which case you'd need to do an element-wise multiply.

specter119:
@computron with pleasure, I have update this PR.

computron:
Thanks!

PR #227, opened 2018-04-27
DB Update #2

shyamd:
Let's try this again.

Updates for DB insertion and Drone:

- switched to a single use_gridfs option for both Bandstructures and DOS
- cleaned up drone auto parsing for electronic structure
- fixed max force causing drone to set state to error (retain the max force analysis)


computron:
Thanks!

PR #241, opened 2018-06-06
QChem Atomate

samblau:
## Summary

QChem comes to Atomate! We have two workflows that run on Cori - one for simulating a dihedral potential and one for doing a double frequency flattening optimization, first in vacuum and then in PCM.

* Firetasks:
   - WriteInputFromIOSet, WriteCustomInput, and WriteInput in firetasks/write_inputs.py
   - RunQChemDirect, RunQChemCustodian, RunNoQChem, and RunQChemFake in firetasks/run_calc.py
   - QChemToDb in firetasks/parse_outputs.py, which depends on QChemDrone in drones.py
   - RotateTorsion in firetasks/geo_transformations.py
   - All firetasks tested in firetasks/tests and drone tested in tests/test_drones.py
* Fireworks:
   - OptimizeFW and FrequencyFlatteningOptimizeFW in fireworks/core.py
   - Both tested in fireworks/tests/test_core.py
* Workflows:
   - get_wf_torsion_potential in workflows/base/torsion_potential.py
   - get_wf_double_FF_opt in workflows/base/double_FF_opt.py
   - Three workflow tests in workflows/tests
* Powerups:
   - use_fake_qchem in powerups.py to enable workflow testing
   - Tested in tests/test_powerups.py

## To Do

* Write documentation
* Write more workflows:
   - Bond dissociation energy workflow
   - Complete basis set extrapolation workflow?
* Discuss other desired functionality
* Review code structure for general improvements
* Improve database.py


computron:
@samblau looks like the tests are not passing due to openbabel installation. Would suggest:

1. Figure out how to get CircleCI to install / use openbabel so it runs the tests properly
2. raise a SkipTest if the user doesn't have the proper installation for Qchem workflows, so if someone downloads atomate for VASP, doesn't install openbabel, and runs the unit tests, it will just skip those tests.

atomate/qchem/database.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def build_indexes(self, indexes=None, background=True):
+        """
+        Build the indexes.
+
+        Args:
+            indexes (list): list of single field indexes to be built.
+            background (bool): Run in the background or not.
+
+        TODO: Figure out what this does
```
computron: Building indexes just makes certain queries faster, it is not required. But if there are certain keys that you want to query on frequently, it's better to add an index so that querying on them is fast.

So I'd suggest just adding a few indices on the common query keys

atomate/qchem/firetasks/write_inputs.py
```diff
@@ -18,32 +27,114 @@ class WriteInputFromIOSet(FiretaskBase):
...
 
-    required_params = ["molecule", "qchem_input_set"]
-    optional_params = ["qchem_input_params"]
+    required_params = ["qchem_input_set"]
+    optional_params = [
+        "molecule", "qchem_input_params", "input_file", "write_to_dir"
+    ]
 
     def run_task(self, fw_spec):
+        input_file = "mol.qin"
```
computron: You can shorten code like this to:

``input_file = self["input_file"] if "input_file" in self else "mol.qin"``

computron: or

``input_file = self.get("input_file", "mol.qin")``

computron: (note that is kind of thing occurs throughout the code)

atomate/qchem/fireworks/tests/test_core.py
```diff
@@ -0,0 +1,191 @@
...
+
+from __future__ import division, print_function, unicode_literals, absolute_import
+
+import os
+import unittest
+
+from atomate.qchem.firetasks.write_inputs import WriteInputFromIOSet
+from atomate.qchem.firetasks.run_calc import RunQChemCustodian
+from atomate.qchem.firetasks.parse_outputs import QChemToDb
+from atomate.qchem.fireworks.core import *
```
computron: avoid import * like this

Better if you can use an IDE to help you add imports automatically. See also:

https://stackoverflow.com/questions/2360724/what-exactly-does-import-import?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

computron:
Nice job!

I only put a few comments here and there. I think it's just a matter of those minor things plus getting the tests to pass on CircleCI.

atomate/qchem/fireworks/tests/test_core.py
samblau: Thanks, fixed

atomate/qchem/firetasks/write_inputs.py
samblau: Thanks, I believe I fixed this everywhere.

atomate/qchem/database.py
samblau: Thanks, added a few single field indices. Left building compound indices as a TODO once we have a better idea of how we'll be using our database.

samblau:
Thanks for the feedback! I've dealt with the minor things. I will be back at codebusters tomorrow and will work on the CircleCI and SkipTest stuff.

PR #242, opened 2018-06-07
update circleci config to actually run tests depending on Mongo

### LINKED ISSUE #204: fix CircleCI tests so that VASP integration tests are also run
It seems the CircleCI integration tests skip most of the tests. Probably this is because running VASP tests (even the simulated VASP runs) requires MongoDB (think this is already set up and so shouldn't be a problem) and also VASP POTCAR files with associated set up of PMG_VASP_PSP_DIR (probably the issue).

ardunn:
## Summary

* Update circleci config.yml to start mongo daemon and run all tests instead of just skipping them
* closes #204 

## TODO (if any)

- [x] see if ~/.pmgrc.yaml + explicit POTCARs fixes ferroelectric test (**it does**)
- [x] are there other POTCARs meant for atomate testing already in atomate? (**appears not, should be incorporated in primary docker image**)
-  [x] figure out why NEB test keeps failing (is it bc bad CircleCI config or something else?) (**moved to issues**)


ardunn:
@computron

PR #249, opened 2018-07-25
Updates to Vasp parsing and db insertion

shyamd:
## Summary

Few small features towards cleaning up passing data from one vasp calculation to another:
- Added the ability to push fields from the task doc in VaspToDB Firetask
- Running and storing Bader information


shyamd:
This is ready to merge. Codacy has one complaint about a multi-line string. I don't have the permissions to ignore that rule for this check.

PR #253, opened 2018-08-08
Refactor QChem file names and update requirements

samblau:
I noticed that my other pull request suddenly included freshly broken tests after merging the latest updates to Atomate and changing the requirement to the newly released Pymatgen version. Thus I wanted to isolate those tests from my fragmentation code that remains under development. I also want to get the QChem refactor complete across the three repos. Thus, this pull request:

- Refactors QChem file names in Pymatgen and Custodian
- Updates the requirements to the latest Pymatgen release and the current Custodian master

We will likely want to wait to merge this until we have a stable Custodian release. But let's see if these few small changes break tests.

### DIFF
--- a/atomate/lammps/drones.py
+++ b/atomate/lammps/drones.py
@@ -11,7 +11,7 @@ from datetime import datetime
 import six
 
 from pymatgen.apps.borg.hive import AbstractDrone
-from pymatgen.io.lammps.output import LammpsLog, LammpsDump, LammpsRun
+# from pymatgen.io.lammps.output import LammpsLog, LammpsDump, LammpsRun
 # from pymatgen.io.lammps.sets import LammpsInputSet
 
 from atomate.utils.utils import get_uri
--- a/atomate/lammps/drones.py
+++ b/atomate/lammps/drones.py
@@ -12,7 +12,7 @@ import six
 
 from pymatgen.apps.borg.hive import AbstractDrone
 from pymatgen.io.lammps.output import LammpsLog, LammpsDump, LammpsRun
-from pymatgen.io.lammps.sets import LammpsInputSet
+# from pymatgen.io.lammps.sets import LammpsInputSet
 
 from atomate.utils.utils import get_uri
 
--- a/atomate/lammps/firetasks/__init__.py
+++ b/atomate/lammps/firetasks/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 
-from .write_inputs import *
-from .run_calc import *
-from .parse_outputs import *
-from .glue_tasks import *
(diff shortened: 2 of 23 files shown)


samblau:
All QChem tests now pass. However, eight tests still error, including the Vasp's TestElasticWorkflow and multiple errors each coming from these two error messages:

ModuleNotFoundError: No module named 'pymatgen.analysis.gb'
ModuleNotFoundError: No module named 'pymatgen.io.lammps.sets'

Seems almost certain that these are due to the latest Pymatgen.

computron:
Thanks!

- Can you remove the "bleeding-edge" custodian requirement? Sometimes atomate lets this slide, but it would be preferable to have a real release being the requirement
- For the failures , I will have to take a look when I have a little more time so that I can fix all the various tests. It might be simple (e.g., just some imports) or it might not. Certain packages like LAMMPs I need to figure out what to do with, I know some people are interested in it but I'm not sure what the future of that module should be.

samblau:
If I remove the bleeding edge custodian requirement, all QChem tests will fail, since the refactor was just applied.

shyuep:
So-called bleeding edge custodian tests fail anyway. Fix those first and I can release.

Shyue Ping

samblau:
@shyuep I have an open Custodian pull request here (https://github.com/materialsproject/custodian/pull/81) which passes all tests

samblau:
I will update the requirements pending the Custodian release. Regarding the remaining failing tests, I'm a bit confused by a data point I got this morning - Evan's pull request into my Atomate repo only has failing LAMMPS tests. I'm hoping I can get this pull request to a similar point tomorrow, and then we could forcibly skip the LAMMPs tests for now to have a clean build. I'll comment again when I get to that point.

shyuep:
Custodian is releaed.

Shyue Ping

samblau:
I have commented out various LAMMPS code to avoid those errors for now. The remaining failing tests are dealt with by switching the Pymatgen requirement from the latest release to the master repo, thus the changes since the recent release are responsible for fixing the tests. Would we prefer to be building against a release, but have failing tests, or build cleanly against the master repo?

montoyjh:
So I think the issue is that the pymatgen.analysis.gb module doesn't have an __init__.py, which means it doesn't get detected as necessary in the pip setup procedure.  I'll file an issue with pymatgen.

montoyjh:
Issued a [PR to pymatgen](https://github.com/materialsproject/pymatgen/pull/1235), hopefully a re-release with that fix will allow pip installs to work properly.

shyuep:
Done.

computron:
Great! Merging so you have it in master and will try to cut a release before the workshop too

PR #254, opened 2018-09-15
NMR Workflows

shyamd:
## Summary
Add an NMR Workflow

* Incorporate MPNMRSet into a WriteVasp firetask
* Build NMRFW that can run a shielding or EFG calculation
* Simple generic NMR workflow that begins with structure opt and incorporates both calculations


computron:
As far as I can tell, there are no tests?

Also I am starting to wish I was less lenient about what goes in fireworks/core.py in the past. Perhaps we can be a bit more organized for the future. Can you move the FW to fireworks/nmr.py?

computron:
@shyamd any plans to add tests soon?

Otherwise I would suggest to close the PR and re-open when ready

shyamd:
Yes, I will try and get to this weekend.

computron:
I made some really minor cleanups

The tests are still weak sauce though ...

Any reason to not do the virtual VASP run and step through the workflow like the other tests?

shyamd:
Added a virtual VASP run test. Most of the real functionality is in pymatgen, but for now, this test ensures everything is working the way the workflow and possibly builders expect it to.

PR #257, opened 2018-09-29
Chgcar storage

jmmshn:
## Summary
* Added storage for CHGCAR type files.
* Added tests for CHGCAR and AECCAR storage
* Changed fireworks/core to allows kwargs to be passed for VaspToDb object

### DIFF
--- a/atomate/vasp/database.py
+++ b/atomate/vasp/database.py
@@ -2,7 +2,7 @@
 
 from __future__ import division, print_function, unicode_literals, absolute_import
 
-from monty.json import MontyEncoder
+from monty.json import MontyEncoder, MontyDecoder
 
 """
 This module defines the database classes.
@@ -14,6 +14,7 @@ from bson import ObjectId
 
 from pymatgen.electronic_structure.bandstructure import BandStructure, BandStructureSymmLine
 from pymatgen.electronic_structure.dos import CompleteDos
+from pymatgen.io.vasp import Chgcar
 
 import gridfs
 from pymongo import ASCENDING, DESCENDING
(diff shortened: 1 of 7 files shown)


computron:
Hi - thanks for the contribution - can you look at the failing py3 test on CircleCI?

jmmshn:
This should work with the next release of pymatgen, where I have made the volumetric objects MSONables

mkhorton:
Update the requirements.txt to require the latest version of pymatgen https://github.com/hackingmaterials/atomate/blob/master/requirements.txt

mkhorton:
@computron any thoughts on adding an additional CI that pulls from pymatgen master? I know a couple of people working on features in pymatgen and atomate concurrently that would benefit from this

computron:
@mkhorton can you open up a separate issue regarding versioning? List out proposed strategy + any pros/cons. I thought @samblau had already created a similar issue but can't seem to find it.

shyuep:
Note that this doesn't really solve the issue. If someone is indeed working on pymatgen, the likelihood is that they need to fork it and then submit a PR when ready. The time lag between pmg master and release is not really that long. It makes more sense for someone to submit a PR for pmg first, wait for it to be accepted, and released, before submitting a PR for atomate.

samblau:
Sorry about that - I got most of the way through researching continuous versioning strategies when I realized that the amount of manual work required to release a new version of pymatgen would preclude any automatic releasing. Additionally, I was informed that having too many frequent releases could have other unintended consequences, at which point I concluded that our current system is probably good enough, despite its limitations. 

@shyamd, @mkhorton and I just talked through it again for a few minutes, and we similarly concluded that automatic releases could introduce too many problems, but that simply having two separate, labeled tests in atomate - one against the latest pymatgen release and the other against the master pymatgen repo - would be the easiest solution to provide both the pull requester and @computron with more complete information about which tests pass in which cases.

Please let me know if you would like me to open an issue about continuous versioning (though technically, we're talking about continuous versioning for pymatgen being the important thing rather than for atomate) or the addition of an additional test against the master pymatgen (and custodian?) repo.

computron:
AFAIK continuous versioning for atomate is not really that important since I don't hear much about people waiting for an atomate release. I'd pause that one.

For the tests against pymatgen master as an additional test, that sounds good to me if somewhat sets that up. It might also just be something to note in the PR (i.e., "wait for pymatgen version update before merge").

computron:
Thanks!

PR #258, opened 2018-10-01
QChem cleanup and fragmenter firetask / firework / workflow

samblau:
## Summary

* Fragmenter firetask added at qchem/firetasks/fragmenter.py. All fragment generation has been moved to pymatgen.analysis.fragmenter, thus this firetask only manages generating molecules of the correct charge and multiplicity from the list of fragments, checking the database to see if a molecule has been previously simulated, and then building new fireworks. Many tests added at qchem/firetasks/tests/test_fragmenter.py.
* FragmentFW added to qchem/fireworks/core.py which essentially just calls the fragmenter firetask. Additional tests added to qchem/fireworks/tests/test_core.py.
* Fragmentation workflow added at qchem/workflows/base/fragmentation.py. Optionally optimizes a principle molecule before calling FragmentFW. Tests added at qchem/workflows/tests/test_fragmentation.py.
* Better handling of single point calculations, which are necessary when a single atom fragment is encountered. SinglePointFW added to qchem/fireworks/core.py and tests added to qchem/fireworks/tests/test_core.py.
* Small changes and cleanup through atomate/qchem as encountered during fragmenter development and based on feedback from the previous pull request. Examples include the drone now parsing the original input file, drone better handling unstable optimizations and failed frequency flattening, max_cores being obtained from the fw_env in all cases, and additional documentation for qchem_input_params in every file where it is used. 

Note that the requirements currently use my pymatgen qchem branch. As usual, merging this PR will need to wait until my open pymatgen PR is merged and a new pymatgen version is released.


computron:
Since I am not so familiar with the QChem code - @espottesmith can you review this PR?

My only comments are some of the input parameters:
1. qchem_cmd - does this support env_chk? If so you can list in docstring?
2. multimode - does this support env_chk? If so you can list in docstring?
3. input_file / output_file - are these options really necessary, i.e., will people want to change them? One reason to exclude them is because it makes it seem (perhaps) like the FW is expecting you to have those files on hand in order to run this Firework, which I don't think is the case (it's just output filenames that you want).

espottesmith:
Yes. I will gladly take a look at this. Have never reviewed before; should I just post comments here, or is there something more that I need to do?

computron:
You can either put comments or do a code review using github tools (show diff then review changes)

@samblau I might have broken your tests by resolving a merge conflict in requirements.txt?

samblau:
Yeah, we still need to be building against pymatgen master rather than a release since all the fragmenter stuff in atomate depends on my latest pymatgen PR. I'll address the points you mentioned above and look forward to hearing Evan's comments.

samblau:
@computron qchem_cmd supports env_chk, I just forgot to say that in the documentation in core.py. However, I also currently give it a default value of "qchem" for all FWs in core, despite the fact that my actual workflows default to >>qchem_cmd<<. Perhaps I should just be consistent and get it from the environment everywhere?

Multimode does not support env_chk, and fundamentally wouldn't make sense to get from the environment since either openMP or MPI could technically be used on any system. Also, in truth, you can actually use both openMP and MPI simultaneously. However, in my experience, QChem runs best on a single node with openMP parallelization over all the cores on that node, which is why I make openMP default. But on the other hand, I didn't want to prevent a user from going the MPI route if they really wanted to. Kinda a weird situation.

You're right about input_file and output_file. I've insulated the user from these parameters at the workflow level, and I should probably do the same at the firework level. It's really only at the firetask level where I can dream up a scenario in which a user would need access to them, for instance if they wanted to use parse_outputs.py to parse a calculation run previously outside of atomate. Will take care of this shortly.

Thanks for your comments!

computron:
Ok great! A few quick follows

- It's fine by me if the default doesn't qchem_cmd use env_chk. No need to complicate things for the general user
- For multimode - one thing we were hoping to do in atomate that we didn't do in MPWorks was to try the separate the concerns of the "science" of the workflow vs the "execution details" - assuming the two are independent. Ideally, most of atomate can just focus on the science of the workflows. Any options regarding execution (like path to qchem cmd, the type of MPI you want to use, queue adapter stuff) would not necessarily be pre-determined ahead of time in the workflow - especially since different machines might have different execution environments, and you may not know ahead of time which machine you're going to run on. Instead, all these details about execution could just be set as env_chk parameter and the person doing the my_fworker.yaml can control the execution mode. Basically, the person deciding which workflows to run might not be the same person who is in charge of executing the workflows or making sure they run, so decisions on execution should get decoupled if possible
- input and output file sounds good!

samblau:
OK, your multimode point makes sense, though at the same time you've convinced me that qchem_cmd actually should default to using env_chk since it's actually even more likely to be set in stone by whatever system you're running on. Working on changes now.

samblau:
That's a good point about it being able to work by default without env_chk. But here's an awkward example - if I leave the default QChem command as "qchem", that won't work by default on NERSC, which requires "qchem -slurm".

computron:
ok - i'd say up to you and @espottesmith on how you want to do it

samblau:
OK, qchem_cmd does now default to >>qchem_cmd<< and multimode defaults to >>multimode<<, but if it isn't passed by the user or present in the environment then it becomes "openmp" so that it will work without env_check for a novice user. I could do the same for qchem_cmd, but I think it's reasonable to just force to user to either define it with their firework or in their environment since it's so important. Same with max_cores, though I could see an argument for having a fallback default of 1 core.

Input file and output file names have been removed as arguments from all fireworks. 

I personally find the idea of making an orig folder to be more clunky than how I have it now, but we could chat more about that over slack.

Since you have the most experience with smd/smx, perhaps you'd want to work on this and do a PR into my repos?

We're on the same page with build_MoleculeGraph.

If you make them top level imports, you end up with an import loop that breaks everything since both files then import each other. Try it out on your local machine - it's pretty funny honestly. 

I've made get_fragmentation_wf have a default of depth=1.

FYI, I'm going to be on vacation Monday and Tuesday, but let's touch base on this again Wednesday.

espottesmith:
I think forcing a user to define their qchem command is reasonable. I don't know how I feel about forcing specification of number of cores. It's probably fine in any reasonable case. I don't think I've ever run a job where I didn't explicitly specify number of cores.

Definitely happy to talk more about the orig implementation. Also would be happy to add the smd/smx functionality to my to-do list.

Makes sense. I figured you had a good reason for the function-level import, but I couldn't see what exactly it was.

Enjoy the vacation. I'll hopefully have PRs ready by the time you get back.

samblau:
Thanks, Evan! Of course, we still need to wait for new pymatgen and custodian versions, but that's how it goes.

computron:
Lets merge now

Few people will be using this anyway at the moment

If you can keep an eye on the custodian release and submit a PR when that's there, that would be helpful.

samblau:
Will do. Thanks again!

PR #259, opened 2018-10-09
Add sanity check to CalcDb.from_db_file

acrutt:
## Summary

Previously no warning was issued if only read_only credentials were defined, leading to password not being set correctly. This leads to a confusing error message where `self.db.counter.find({"_id": "taskid"}).count()` fails due to an authentication error.

## TODO (if any)

None

### DIFF
--- a/atomate/utils/database.py
+++ b/atomate/utils/database.py
@@ -108,6 +108,12 @@ class CalcDb(six.with_metaclass(ABCMeta)):
             MMDb object
         """
         creds = loadfn(db_file)
+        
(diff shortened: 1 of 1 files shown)


mkhorton:
Looks like we may have to change this solution after all, since this solution won't work for databases which don't use authentication at all (i.e. the local test database). Hmm.

Perhaps changing:

`if admin and "admin_user" not in creds`

to:

`if admin and "admin_user" not in creds and "readonly_user" in creds` ?

This seems a little brittle though, maybe there's a better way to fix this.

PR #260, opened 2018-11-01
add AmsetFW which has new RunAmset and AmsetToDb tasks

albalu:
## Summary

* I used AmsetFW by adding it after a "nscf uniform" firework. It uses the vasprun.xml from this calculation similar to BoltztrapFW; it also has RunAmset and AmsetToDb very similar to BoltztrapFW 
* AmsetToDb: the mobility and seebeck values are stored as list of jsons that contain the carrier concentration (c) and the temperature (T) at which the property is calculated as well as the actual vector ("vec") and average ("avg") of the calculated vector.
* unlike BoltzTrapFW, Amset is not under pymatgen and is not and should not be an atomate requirement. Therefore, the way I addressed this is to import Amset right under run_task as opposed to the top of the file (hence those who don't have Amset installed won't get ImportError).

### DIFF
--- a/atomate/vasp/firetasks/parse_outputs.py
+++ b/atomate/vasp/firetasks/parse_outputs.py
@@ -269,14 +269,18 @@ class AmsetToDb(FiretaskBase):
     Optional params:
         db_file (str): path to file containing the database credentials.
             Supports env_chk. Default: write data to JSON file.
-        additional_fields (dict): fields added to the document such as user-defined tags or name, ids, etc
+        filename (str): the json filename from which the Amset run is loaded
+            via the from_file method.
+        additional_fields (dict): fields added to the document such as
+            user-defined tags or name, ids, etc
     """
 
     optional_params = ["db_file", "filename", "additional_fields"]
 
     def run_task(self, fw_spec):
         from amset.core import Amset
-        amset = Amset.from_file()
+        filename = self.get("filename", "amsetrun.json")
+        amset = Amset.from_file(filename=filename)
         additional_fields = self.get("additional_fields", {})
 
         # pass the additional_fields first to avoid overriding BoltztrapAnalyzer items
--- a/atomate/vasp/firetasks/parse_outputs.py
+++ b/atomate/vasp/firetasks/parse_outputs.py
@@ -335,12 +335,11 @@ class AmsetToDb(FiretaskBase):
             amset_dict (dict): the output of Amset.from_file() method
 
         Returns (({[dict]}, [dict]):
-            For mobility (the first output): it is a dictionary containing
-                list of dicts for mobility vectors limited by various types of
-                scattering phenomena and the "overall" mobility at various
-                carrier concentration, c, and temperatures, T. The second
-                output is a list of dicts for seebeck coefficient also at
-                various c and T.
+            The first output (mobility) is a dictionary containing list of
+            dicts for mobility vectors limited by various types of scattering
+            phenomena and the "overall" mobility at various carrier
+            concentration, c, and temperatures, T. The second output is a list
+            of dicts for seebeck coefficient also at various c and T.
         """
         mobility = {}
         seebeck = []
--- a/atomate/vasp/firetasks/parse_outputs.py
+++ b/atomate/vasp/firetasks/parse_outputs.py
@@ -343,7 +343,7 @@ class AmsetToDb(FiretaskBase):
                             "vec": d["seebeck"][tp][c][T],
                             "avg": np.mean(d["seebeck"][tp][c][T])
                         })
-        
+
         d["mobility"] = mobility
         d["seebeck"] = seebeck
         db_file = env_chk(self.get('db_file'), fw_spec)
--- a/atomate/vasp/firetasks/parse_outputs.py
+++ b/atomate/vasp/firetasks/parse_outputs.py
@@ -317,6 +317,7 @@ class AmsetToDb(FiretaskBase):
         d["tags"] = list(set(d.get("tags", []) + fw_spec.get("tags", [])))
 
         # transform mobility and Seebeck coefficient to list of dicts:
+        # TODO: move this part to Amset for clearer input/output
         mobility = {}
         seebeck = []
         for itp, tp in enumerate(["p", "n"]):
@@ -342,6 +343,7 @@ class AmsetToDb(FiretaskBase):
                             "vec": d["seebeck"][tp][c][T],
                             "avg": np.mean(d["seebeck"][tp][c][T])
                         })
+        
         d["mobility"] = mobility
         d["seebeck"] = seebeck
         db_file = env_chk(self.get('db_file'), fw_spec)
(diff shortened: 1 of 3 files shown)


atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -259,6 +259,93 @@ def run_task(self, fw_spec):
...
+@explicit_serialize
+class AmsetToDb(FiretaskBase):
+    """
+    Enter an Amset run into the database. Note that this assumes you are in a
+    current directory that has the data with a sub-directory called "run_data"
+    containing the "amsetrun.json.gz" file (Amset run default filename).
+
+    Optional params:
+        db_file (str): path to file containing the database credentials.
+            Supports env_chk. Default: write data to JSON file.
```
computron: filename parameter is not documented or even used!

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -259,6 +259,93 @@ def run_task(self, fw_spec):
...
+        additional_fields (dict): fields added to the document such as user-defined tags or name, ids, etc
+    """
+
+    optional_params = ["db_file", "filename", "additional_fields"]
+
+    def run_task(self, fw_spec):
+        from amset.core import Amset
+        amset = Amset.from_file()
```
computron: confusing that this is an empty arg ...

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -259,6 +259,93 @@ def run_task(self, fw_spec):
...
+        d["spacegroup"] = {"symbol": sg.get_space_group_symbol(),
+                           "number": sg.get_space_group_number(),
+                           "point_group": sg.get_point_group_symbol(),
+                           "source": "spglib",
+                           "crystal_system": sg.get_crystal_system(),
+                           "hall": sg.get_hall()}
+        d["created_at"] = datetime.utcnow()
+        d["tags"] = list(set(d.get("tags", []) + fw_spec.get("tags", [])))
+
+        # transform mobility and Seebeck coefficient to list of dicts:
```
computron: might be better to move this into an internal function so that run_task stays relatively clean and high level, and so the documentation for the function has clearly defined input and output formats

atomate/vasp/firetasks/run_calc.py
```diff
@@ -241,6 +242,60 @@ def run_task(self, fw_spec):
         runner.run(path_dir=os.getcwd())
 
 
+@explicit_serialize
+class RunAmset(FiretaskBase):
+    """
+    Run Amset directly. Requires vasprun.xml to be in current dir.
```
computron: There should be an option to give path to the vasprun.xml file

atomate/vasp/firetasks/run_calc.py
```diff
@@ -241,6 +242,60 @@ def run_task(self, fw_spec):
...
+
+    Required params:
+        material_params (dict): the static dielectric constant (epsilon_s) is
```
computron: Tell people where to find more information about what goes inside this dictionary. Very unclear what else can go in and what happens if more or less info is specified

atomate/vasp/firetasks/run_calc.py
```diff
@@ -241,6 +242,60 @@ def run_task(self, fw_spec):
...
+            mandatory.
+
+    Optional params:
+        model_params (dict): parameters related to the model used and the level
+            of theory. See Amset documentation for key options.
+        performance_params (dict): parameters related to convergence, speed,
+            etc. See Amset documentation for key options
+        temperatures ([float]): the temperatures (in Kelvin)
+        doping: ([float]) doping levels you want to compute
```
computron: document all optional_params, including kgrid_tp, write_outputs, write_inputs

atomate/vasp/firetasks/run_calc.py
```diff
@@ -241,6 +242,60 @@ def run_task(self, fw_spec):
...
+    """
+    required_params = ["material_params"]
+    optional_params = ["model_params", "performance_params",
+                       "temperatures", "dopings", "coeff_file",
+                       "kgrid_tp", "write_outputs", "write_inputs"]
+
+    def run_task(self, fw_spec):
+        from amset.core import Amset
+        material_params = self.get("material_params")
+        model_params = self.get("model_params", None)
```
computron: why sometimes None and sometimes ``{}`` (as in the next line)

atomate/vasp/firetasks/run_calc.py
```diff
@@ -241,6 +242,60 @@ def run_task(self, fw_spec):
...
+        perfs = self.get("performance_params", {})
+        perfs["max_nvalleys"] = perfs.get("max_nvalleys", 3)
```
computron: make sure the defaults work well for most materials. Although ideally people should fully read the Amset documentation and set these parameters intelligently, the reality is that 90% of people will just run the AmsetFW without thinking about it and report that as the "amset result". So make sure that these defaults give good numbers as they will be representative of what people report as the Amset value.

atomate/vasp/firetasks/run_calc.py
```diff
@@ -241,6 +242,60 @@ def run_task(self, fw_spec):
...
+        perfs["max_nbands"] = perfs.get("max_nbands", 1)
+        temperatures = self.get("temperatures", None)
+        dopings = self.get("dopings", None)
+        coeff_file = self.get("coeff_file", None)
+        kgrid_tp = self.get("kgrid_tp", "coarse")
+        write_outputs = self.get("write_outputs", True)
+        write_inputs = self.get("write_inputs", True)
+        timeout = self.get("timeout", 10)
```
computron: is this hours? This is an unconventional unit for timeout (usually minutes or seconds). Might want to call the parameter ``timeout_hours``.

atomate/vasp/firetasks/run_calc.py
computron: (unless this is a BoltzTrap limitation, in which case maybe write in the docs that this is the reason)

albalu:
Thanks @computron , I addressed all of these comments except moving this part in AmsetToDb to Amset itself: "transform mobility and Seebeck coefficient to list of dicts" 
Instead I added a todo. The reasons is that I won't be able to properly test the result of this moving in a short amount of time.

computron:
Ok although I meant just having it be an internal function of the same class, not moving it to the amset codebase (unless it's useful there). The former should hopefully just take a few moments, it's just copy pasting the code into its own function in that class and creating a docstring for that function.

computron:
(also the unit tests)

albalu:
ah sorry I misunderstood the internal function. I fixed that. Regarding unit-tests, it would be great if we could defer that until Amset is pip installable and this FW gets used by more people than just myself so we find at least one edge case.

computron:
sorry, no unit test, no merge into master

computron:
This PR is quite stale and likely will be rewritten from scratch given all the various AMSET updates.

I am closing this.

@utf FYI in case you find any of this code useful

PR #266, opened 2019-01-14
[Merge conflicts, WIP] Magnetism deformation and ordering workflows

mkhorton:
See #181 for more discussion.


computron:
@mkhorton can you look at the Codacy warnings

For things like "unused variable v" consider replacing with a standard dummy variable ("_")

atomate/vasp/workflows/base/magnetism.py
```diff
@@ -0,0 +1,834 @@
...
+               those attempted in this specific workflow). For
+               high-throughput studies, a dedicated builder is
+               recommended.
+            5. For book-keeping, a record is kept of whether the
+               input structure is enumerated by the algorithm or
+               not. This is useful when supplying the workflow with
+               a magnetic structure obtained by experiment, to measure
+               the performance of the workflow.
+
+        :param structure: input structure
```
computron: Use same docstyle as rest of atomate (and pymatgen, etc.)

atomate/vasp/workflows/base/magnetism.py
```diff
@@ -0,0 +1,834 @@
...
+            )
+        )
+
+        # when enumerating ferrimagnetic structures, it's useful to detect
+        # symmetrically distinct magnetic sites, since different
+        # local environments can result in different magnetic order
+        # (e.g. inverse spinels)
+        # initially, this was done by co-ordination number, but is
+        # now done by a full symmetry analysis
+        sga = SpacegroupAnalyzer(structure)
```
computron: Some of this code regarding finding unique sites for magnetization - why put in atomate rather than pymatgen?

In general I see pymatgen as the place for doing any algorithm development, atomate should mostly be calling pymatgen with some settings for high-throughput workflows.

computron: especially since (as far as I can tell) your input and output objects are not workflows, fireworks etc, this is just structure enumeration

atomate/vasp/workflows/base/magnetism.py
```diff
@@ -0,0 +1,834 @@
...
+
+        # ... and strip out existing magmoms, which can cause conflicts
+        # with later transformations otherwise since sites would end up
+        # with both magmom site properties and Specie spins defined
+        if "magmom" in input_structure.site_properties:
+            input_structure.remove_site_property("magmom")
+
+        return input_structure
+
+    def _generate_transformations(self, structure):
```
computron: I know this is an internal method, but better docstring about input and output parameters would help

atomate/vasp/workflows/base/magnetism.py
```diff
@@ -0,0 +1,834 @@
...
+        # and then use each to perform a transformation later
+        all_constraints = {}
+
+        # ...to which we can add simple AFM cases first...
+        if "antiferromagnetic" in self.strategies:
+
+            constraint = MagOrderParameterConstraint(
+                0.5,
+                # TODO: this list(map(str...)) can probably be removed
+                species_constraints=list(map(str, types_mag_species)),
```
computron: can you remove list(map(str ?

If you leave it as a comment, no one is going to ever touch it ...

atomate/vasp/workflows/base/magnetism.py
```diff
@@ -0,0 +1,834 @@
...
+                mag_species_spin,
+                order_parameter=constraints,
+                **self.transformation_kwargs
+            )
+
+            transformations[name] = trans
+
+        return transformations
+
+    def _generate_ordered_structures(self, sanitized_input_structure, transformations):
```
computron: See my comment regarding the previous function and pymatgen vs atomate

Ideally, atomate routines should be short and just glue together existing functionality in pymatgen into a set of Fireworks and Workflow objects. i.e. call a bunch of pymatgen functions and wrap that into a Firework/Workflow object.

When a lot of algorithm work is put in atomate, two problems occur:
* As atomate is less used than pymatgen, fewer people and developers review, vet, and improve the algorithms
* Generally, tying together workflows is confusing enough code for people to follow as workflows are somehow inherently confusing. When you mix in algorithm work together with the workflow stuff (even if the algorithms are in a separate private function), wrapping your head around workflows becomes even harder. So to keep atomate workflows easy to maintain, it's best if they are as short as possible and use pymatgen for most of the heavy lifting

atomate/vasp/workflows/base/magnetism.py
```diff
@@ -0,0 +1,834 @@
...
+        """
+        Apply our input structure to our list of transformations and output a list
+        of ordered structures that have been pruned for duplicates and for those
+        with low symmetry (optional).
+        """
```
computron: better docstring for input/outputs (even for private function)

atomate/vasp/workflows/base/magnetism.py
```diff
@@ -0,0 +1,834 @@
...
+
+        return
+
+    def get_wf(
+        self, scan=False, perform_bader=True, num_orderings_hard_limit=16, c=None
+    ):
+        """
+        Retrieve the FireWorks workflow.
+
+        :param scan: if True, use the SCAN functional instead of GGA+U, since
```
computron: docstring format (see previous comment)

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -753,6 +755,300 @@ def run_task(self, fw_spec):
         logger.info("Thermal expansion coefficient calculation complete.")
 
 
+@explicit_serialize
+class MagneticOrderingsToDB(FiretaskBase):
```
computron: We should talk about the "builder" vs "firework" way to do this

Honestly I think builders should be used for essentially all these problems. The strategy of running a single Firework once at the end of a workflow, and assuming things are not going to change later, is just going to cause problems down the line.

Maybe schedule a phone call to discuss?

atomate/vasp/workflows/base/magnetism.py
mkhorton: Yes, it's definitely a balance. I also commit a lot of improvements to the `MagOrderingTransformation` in pymatgen and a new `MagOrderParameterConstraint` class too. The balance I ended up falling on was "transformation stuff goes into pymatgen, but *choice* of transformation kwargs goes into the workflow" -- and that is what this extra stuff is.

The design of this workflow is a little different in one respect though, there's the class itself, and then a `.get_wf()` method -- the reason for this choice was that I wanted more introspection into the workflow generation, and I also wanted to separate different components out for readability/ease of maintenance.

atomate/vasp/workflows/base/magnetism.py
mkhorton: Noted 👍

atomate/vasp/workflows/base/magnetism.py
mkhorton: Ok, will add

atomate/vasp/workflows/base/magnetism.py
mkhorton: Updated comment above about this. I did add a lot of underlying improvements to pymatgen too, it's just that some extra glue was needed to generate a plausible set of inputs to these updated transformations, and it *felt* like this was more a workflow thing than a pymatgen thing, but that is a judgement call.

atomate/vasp/firetasks/parse_outputs.py
mkhorton: Yes, I was following the previous examples in atomate for this.

My judgement was that the atomate workflow should work as a stand-alone workflow, which is what this task is for: if someone wants to just understand magnetic orderings they can run the workflow and get a nice collection as output.

However, for MP, we will need a builder regardless, and there does seem to be some duplication of effort here. Happy to discuss further!

Issue #262, opened 2018-11-14
Wrong saving of "custodian.json"


gpetretto:
Saving the `custodian.json` to the `stored_data` at the end of a `RunVaspCustodian` task does not work as expected. `stored_data` should be a dict, while `custodian.json` contains a list.
Due to how fireworks handles the stored data in `rocket`
```python
all_stored_data.update(m_action.stored_data)
```
currently in the launch document a `{"job": "correction"}` dictionary is saved.

computron:
Thanks for the report!

What do you think about changing it to:

{"custodian": [[contents of custodian array]]}

gpetretto:
This seems a good solution to me

computron:
I am closing this after the fix

PR #270, opened 2019-02-19
Fix custodian stored data (From issue #262)

sivonxay:
I was having errors with fireworks fizzling due to incorrect saving of the custodian.json.

Looks like the error was brought up in issue #262, but not corrected. This seems to be the agreed upon fix from the discussions.


computron:
Thanks!

PR #276, opened 2019-03-24
Small update for the writting and retrieving of AECCARs

jmmshn:
## Summary
* Added support for AECCARs in the retrieve_task function 
* While running on CORI I noticed that 5-10% of the static calculations produced AECCARS that contained strangely noise data and negative values.  The problem tends to fix itself once the calculations are reran.   I'm not sure how to fix this exactly maybe we need to add a check in custodian if we plan on using the AECCARs for more things.  For now, I will just add a check for positive definite AECCAR0+AECCAR2 in the writing and retrieving routines.


mkhorton:
Check should be added to Custodian also. This is something that needs to get investigated further.

jmmshn:
I'm getting this error on circleCI:
  ```
File "/root/miniconda3/envs/test_env/lib/python3.6/site-packages/pymatgen/io/vasp/sets.py", line 29, in <module>
    from pymatgen.analysis.structure_matcher import StructureMatcher
  File "/root/miniconda3/envs/test_env/lib/python3.6/site-packages/pymatgen/analysis/structure_matcher.py", line 25, in <module>
    from pymatgen.optimization.linear_assignment import LinearAssignment
ImportError: cannot import name 'LinearAssignment'
```
is there a problem with pymatgen install?

mkhorton:
Hmm, nothing new that should have broken it to my knowledge, but someone else also experienced this on CI on atomate specifically. There's an open issue on the linear assignment module to improve its speed, I keep meaning to make all the compiled-optimizations in pymatgen strictly optional so that a pure Python version will still work.

What numpy version is on atomate? Might be inconsistent with the pymatgen version, especially if the CI install is using the wheel.

PR #284, opened 2019-05-03
add powerup approach to gzip changing of RunVaspCustodian

dbroberg:
Simple power-up added for turning gzip off for all RunVaspCustodians in a workflow. This is a necessary requirement for using the get_wf_ferroelectric workflow.


shyuep:
Just a note that all these seem to be a very convoluted way of changing a simple behavior of custodian (which is controlled via a single kwarg). 

In general, I think "power-ups" should really be used for things that truly enhance the workings in some way, rather than turning gzip off. 

I have no idea why the ferroelectric wf requires gzip to be off.

dbroberg:
I didn't add any unit tests because it didn't seem like the other simple powerups had any either.

dbroberg:
@shyuep , Sorry - I didn't see your previous comment.  I would defer to @blondegeek for any comments on the original design of the workflow. I still don't fully understand why turning off gunzip is required, but I believe it has to do with the workflow's interpolation process and not being able to read gunzip'd files at certain stages.

If you would rather, we can just add the two lines included in the powerup to the bottom of the get_wf_ferroelectric function. Just thought the original solution was more inline with what was desired for powerup functions.

shyuep:
Unless the interpolation relies on external codes that do not support gzipped files, I would say it is easier to make the other workflow transparently support gzip as per all other parts of pymatgen. Anyway, I will defer to @computron .

blondegeek:
Do we know which parts of the workflow are failing due to gzip? If we are
to update those portions — how big are the changes? The interpolation step
involves reading two CONTCARs, converting them into Structure objects and
then performing the interpolation. It is also possible that there are other
parsers that are not gzip capable.

mkhorton:
Pretty sure all the parsers are gzip-capable, and if not they can be trivially made to be, gzip shouldn't be a big issue.

shyuep:
The philosophy of PMG is that output files tend to be large and in all cases, monty.io.zopen is used to handle gzip (or bzip) transparently. I would be very surprised if there is some special handling needed. Even if one of the parsers does not work with gzip, it is trivial to do this.

dbroberg:
I believe the issue has to do with CopyFilesFromCalcLoc being fed the filename CONTCAR, rather than CONTCAR.gz... 

If this is the case then this is the order with which it is buried: get_wf_ferroelectric -> LcalcpolFW -> StaticInterpolateFW -> WriteVaspFromIOSetFromInterpolatedPOSCAR -> WriteVaspFromIOSetFromInterpolatedPOSCAR -> CopyFilesFromCalcLoc. 

The failure occurs within interpolate_poscar. Not sure what the best fix is. Is there a simple way for GetInterpolatedPOSCAR to know if the files are gzipped or not?

computron:
Note - I don't agree with turning off gzip as a solution to "fixing" the ferroelectric workflow. As @shyuep said, all the pymatgen/monty/etc. codes are designed to make it very easy (if you know what you are doing...) to support gzipped files. So we definitely don't want any solution where the ferroelectric workflow just turns off gzip automatically just so that it works.

That said, sometimes people want to turn off gzip for other reasons. Maybe they have some other VASP post-processing code or a text editor that doesn't support gzipped files and just don't want this. I think it's fine to have this powerup here for that reason.

So, I am going to merge the part about the powerup, but remove the part about wf_ferroelectric automatically turning off gzip.

As for auto-detecting if a file is gzipped, typically you want to wrap your file path around monty.zpath. e.g.

```
s1 = Structure.from_file(os.path.join(interpolate_folder, "CONTCAR_0"))
```

becomes

```
s1 = Structure.from_file(monty.zpath(os.path.join(interpolate_folder, "CONTCAR_0")))
```

blondegeek:
@samack26  -- Could you make change that @computron gave up above (wrap `os.path.join` with `monty.zpath`) to the ferroelectric workflow on your current branch?

Issue #283, opened 2019-04-25
nscf-uniform calculation and data query in MgO band structure tutorial 


zezhong-zhang:
I was following the MgO band structure [tutorial](https://atomate.org/running_workflows.html). The structure optimisation, statistic calculation and non-self-consistent-line calculations are fine. But for nscf-uniform calculation, vasp fails namelessly as shown in the fig and custodian cannot handle. As suggested by Steven on the google group, it may due to the kpoints generated not matching well with the tetrahedron method ISMEAR=-5.
```python  
# Create a bandstructure from the workflow from the atomate presets
import numpy as np
from pymatgen import Structure
from fireworks import LaunchPad
from atomate.vasp.workflows.presets.core import wf_bandstructure
from atomate.vasp.powerups import add_modify_incar

# load structure from file
struct = Structure.from_file('POSCAR')  # note: many file formats supported, see function docs

# create the Workflow
wf = wf_bandstructure(struct)

# finally, instatiate the LaunchPad and add the workflow to it
lpad = LaunchPad.auto_load() # loads this based on the FireWorks configuration
lpad.add_wf(wf)
```
>Warning & Error Message
WARNING atomate.vasp.drones Transformations file does not exist.
Traceback (most recent call last):
  File "/home/zzhang/atomate/atomate_env/lib/python3.6/site-packages/custodian/custodian.py", line 320, in run
    self._run_job(job_n, job)
  File "/home/zzhang/atomate/atomate_env/lib/python3.6/site-packages/custodian/custodian.py", line 428, in _run_job
    raise CustodianError(s, True, v)
custodian.custodian.CustodianError: (CustodianError(...), 'Validation failed: <custodian.vasp.validators.VasprunXMLValidator object at 0x2aaabc54d2b0>')
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/zzhang/atomate/atomate_env/lib/python3.6/site-packages/fireworks/core/rocket.py", line 262, in run
    m_action = t.run_task(my_spec)
  File "/home/zzhang/atomate/atomate_env/lib/python3.6/site-packages/atomate/vasp/firetasks/run_calc.py", line 205, in run_task
    c.run()
  File "/home/zzhang/atomate/atomate_env/lib/python3.6/site-packages/custodian/custodian.py", line 330, in run
    .format(self.total_errors, ex))
[...]

Issue #282, opened 2019-04-13
Slabs in adsorption wfs are slightly different from slabs in test_files


JosephMontoya-TRI:
Working with @oxana-a on refining the adsorption workflow a bit and I noticed the unit tests are failing with the latest pymatgen release.  I'm not 100% sure, but I believe this to be a result of the way oriented unit cells are made primitive.  As far as I can tell, the results are not meaningfully different, but I want to verify a few things before changing the test files.  Specifically, I want to make sure the SlabTransformation pipeline is still compatible with the strategy we've used for adding adsorbates.

@richardtran415 - have you encountered anything like this?  Specifically, one weird thing about the current unit tests is that a SlabTransformation generated from a given slab's oriented unit cell and shift is different from the slab by a reflection through the midpoint of the c-lattice plane, see the following MWE:

```
from pymatgen import Structure, Lattice
from pymatgen.core.surface import generate_all_slabs
from pymatgen.transformations.advanced_transformations import SlabTransformation

struct_ir = Structure.from_spacegroup("Fm-3m", Lattice.cubic(3.875728), ["Ir"], [[0, 0, 0]])
sgp = {"max_index": 1, "min_slab_size": 7.0, "min_vacuum_size": 20.0}
slabs = generate_all_slabs(struct_ir, **sgp)
slab_100 = [slab for slab in slabs
            if slab.miller_index==(1, 0, 0)][0]

# Generate a slab transformation of the oriented unit cell
trans_params = {"miller_index": [0, 0, 1],
                "shift": slab_100.shift,
                "min_vacuum_size": 20.0,
                "min_slab_size": 7.0}
trans = SlabTransformation(**trans_params)
new_slab = trans.apply_transformation(slab_100.oriented_unit_cell)

print("Old slab:", slab_100.frac_coords)
print("New slab:", new_slab.frac_coords)
```

```
Old slab: [[0.      0.      0.90625]
 [0.5     0.5     0.96875]
 [0.      0.      0.78125]
 [0.5     0.5     0.84375]]
New slab: [[0.      0.      0.09375]
 [0.5     0.5     0.03125]
 [0.      0.      0.21875]
 [0.5     0.5     0.15625]]
```

richardtran415:
[...]

PR #285, opened 2019-05-03
Update boltztrap workflow

utf:
## Summary

As discussed in materialsproject/pymatgen#1458, #283, and materialsproject/pymatgen#1453, uniform NSCF calculations have been failing due to an update in `MPNonSCFSet` (tetrahedron smearing was turned on).

The underlying causes has been fixed in pymatgen. As part of the fix `MPNonSCFSet` now has an additional "boltztrap" mode, which ensures a k-point mesh with the necessary symmetry for boltztrap is used. This PR updates the `bandstructure_boltztrap` workflow to use the new mode.


utf:
The failing tests are related to the adsorbate workflow and are unrelated to the changes made in this PR. See #282 for more details. Otherwise, this should be ready to merge.

PR #288, opened 2019-05-23
Change default vasp_cmd and db_file to those in config.py

mkhorton:
This just updates those in core, I haven't changed these elsewhere.

Note this changes the default db_file from `None` to the value currently specified in config.py (`>>db_file<<`), if we want to keep the default as `None` I would suggest also changing the value in config.py, however I think this is fine as-is.

### DIFF
--- a/atomate/vasp/fireworks/core.py
+++ b/atomate/vasp/fireworks/core.py
@@ -5,7 +5,8 @@ from __future__ import absolute_import, division, print_function, \
 
 import warnings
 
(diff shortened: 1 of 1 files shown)


mkhorton:
Perhaps worth changing env_chk to strict=False by default too?

computron:
Thanks!

Where would you want to default env_chk strict to False? I would think that True would help catch unintentional coding / setup errors

PR #298, opened 2019-06-19
CalcDB from_file method now supports SSL

stevetorr:
## Summary
This pull request is for a bug which I identified and patched in this google groups link:
https://groups.google.com/forum/#!topic/atomate/7D8Cc4xyiLw

Briefly summarized, the issue is: (a full explanation at the link above)
- The from_db method in the CalcDB method cannot take in arguments beyond what is explicitly listed in the function.
- This interferes with users who are using the CalcDB method as a part of workflows which ingest the results of calculations into databases which e.g. require SSL to enter.
And my fix is:
+ I have updated the method to improve flexibility with input SSL parameters.
+ I anticipated and addressed a potential bug in which the kwargs passed to the MongoClient instantiation collide with other arguments already hard-coded to be passed in, and so I remove the keys in the kwargs dictionary which produce a 'collision'; this also solves the problem of unexpected kwargs which could raise an error.Therefore, I decided a safer way to allow SSL support was through a 'whitelist', which would be less likely to break people's installations.

As I mentioned before, I don't know if unsolicited pull requests are welcome for tiny bug fixes, but hopefully you can accept this contribution.

### DIFF
--- a/atomate/utils/database.py
+++ b/atomate/utils/database.py
@@ -135,7 +135,7 @@ class CalcDb(six.with_metaclass(ABCMeta)):
 
         # Remove keys which may cause collision when calling MongoClient
         for key in list(kwargs.keys()):
-            if key not in ['ssl','ssl_ca_certs','ssl_pem_passphrase','ssl_keyfile']:
+            if key not in ['ssl','ssl_ca_certs','ssl_ca_certfile','ssl_pem_passphrase','ssl_keyfile']:
                 del kwargs[key]
         
         return cls(creds["host"], int(creds["port"]), creds["database"], creds["collection"],
--- a/atomate/utils/database.py
+++ b/atomate/utils/database.py
@@ -125,11 +125,19 @@ class CalcDb(six.with_metaclass(ABCMeta)):
             user = creds.get("readonly_user")
             password = creds.get("readonly_password")
 
-        kwargs = {}
+        kwargs = dict(creds)
         if "authsource" in creds:
             kwargs["authsource"] = creds["authsource"]
         else:
             kwargs["authsource"] = creds["database"]
+        if 'ssl' in creds:
+            kwargs['ssl'] = bool(creds['ssl'])
+
+        # Remove keys which may cause collision when calling MongoClient
+        for key in ['host','port','database','collection','admin_password','admin_user',
+               'readonly_user','readonly_password','password','username']:
+            if key in kwargs:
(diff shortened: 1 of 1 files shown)


atomate/utils/database.py
```diff
@@ -125,11 +125,18 @@ def from_db_file(cls, db_file, admin=True):
...
         if "authsource" in creds:
             kwargs["authsource"] = creds["authsource"]
         else:
             kwargs["authsource"] = creds["database"]
-
+        if 'ssl' in creds:
+            kwargs['ssl'] = bool(creds['ssl'])
+
+        # Remove keys which may cause collision when calling MongoClient
+        for key in list(kwargs.keys()):
```
computron: Could you explain the rationale for this?

Are you trying to avoid a specific problem that you've seen in practice happen, or are you simply trying to be extra cautious?

I would generally avoid this; for example, let's say that MongoClient in a later version of pymongo supports some new kwarg based on some new feature of MongoDB. A user supplies that new kwarg, but it fails due to this line of code. The user is then confused - is atomate restricting the keys for some scientific reason (e.g., atomate will break if the new kwarg is supplied) or will it be OK to add the new kwarg to the list of allowed kwargs?

Let me know if there's some specific reason this line needs to be in there, thanks!

computron: Nevermind I see the point now - you want to avoid having things like ``readonly_user`` and ``readonly_password`` from creds passed into kwargs. I think there should be a better solution to this though

computron:
@stevetorr thanks for this PR. In general, yes, atomate of course accepts pull requests! However, if  you are undertaking something time-consuming, it is always better to coordinate with us first to make sure your PR can be put in the main branch without issues.

For this particular issue, I am closing it via: dfebb7fefd380afb0bbcd0e3f48efb1670b1019a

Essentially, the commit above allows you to create a key called ``mongoclient_kwargs`` in your creds file. The subkeys for that (like ssl) can be used to throw any kwarg you want into MongoClient. Note that casting to boolean should not be needed ; both JSON and YAML should support boolean as native types. Let me know if you run into issues with this solution! Here is some documentation on nested dicts in YAML if you need it:

https://stackoverflow.com/questions/12468806/yaml-multi-nested-and-python-dictionary

e.g.
```
mongoclient_kwargs:
    ssl: True 
```

or
```
mongoclient_kwargs: {ssl: True}
```

PR #302, opened 2019-08-26
Update to glue_tasks.py CopyVaspOutputs

tboland1:
## Summary
When using CopyVaspOutputs to copy very large files, like the WAVECAR, current implementation can fail after running out of memory. This is most likely from the way the file is read into the memory with *.read(). Using shutil.copyfileobj which has chunk size optimization avoids this failure and is in general faster.

*Capable of unzipping large file
* Faster
* Fix: Errors which result from copying large files


computron:
Thanks!

Issue #304, opened 2019-08-27
SCAN workflow broken, extraneous Custodian arg?


mkhorton:
"db_file" doesn't seem to be used by RunVaspCustodian?

https://github.com/hackingmaterials/atomate/blob/ded674be240ceb7af097bd95610d2c50120c428e/atomate/vasp/workflows/presets/scan.py#L41

https://github.com/hackingmaterials/atomate/blob/ded674be240ceb7af097bd95610d2c50120c428e/atomate/vasp/firetasks/run_calc.py#L68

Not sure if this changed (@shyamd ?), or if it never did anything, but since strict checks were added the workflow can no longer be initialized.

PR #317, opened 2019-10-25
Fix bug in SCAN workflow

rkingsbury:
## Summary

RunVaspCustodian no longer accepts a db_file argument, so the wf_scan_opt workflow fails. This commit removes the spurious argument.

### DIFF
--- a/atomate/vasp/powerups.py
+++ b/atomate/vasp/powerups.py
@@ -292,28 +292,6 @@ def clear_modify(original_wf, fw_name_constraint=None):
         original_wf.fws[idx_fw].tasks.pop(idx_t)
     return original_wf
 
-def set_walltime(original_wf, walltime=None, 
-                          fw_name_constraint=None, task_name_constraint=None):
-    """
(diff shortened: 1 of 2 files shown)


mkhorton:
Fun fact, to my knowledge it never accepted a db_file arg so this didn’t
actually do anything. This is one of the many little bugs picked up by
stricter checking of FW spec args.

mkhorton:
Would close #304

rkingsbury:
I'm not sure why the tests are failing. The error is 

`ImportError: cannot import name 'VaspInput' from 'pymatgen.io.vasp' (/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymatgen/io/vasp/__init__.py)`

which is unrelated to the code I changed

utf:
Hi Martin,

The atomate unit tests have now been fixed in the master branch.

Would you be able to pull the latest changes from master and update this branch so we can check no tests are failing.

Best,
Alex

rkingsbury:
@utf We're revamping the SCAN workflows in more detail now, so I'm going to close this and resubmit a new PR after the more substantive changes are finalized

PR #319, opened 2019-11-14
Add a warning for user if ISIF 0, 1, 2, or 7 and double relaxation job

mhsiron:
A double relaxation job is typically used when allowing the cell shape and volume to change, in order to avoid Pulay stresses.

I added a warning that if these degrees of freedom aren't allowed but a double relaxation job is selected, perhaps computational time might be wasted.


utf:
Hi Martin,

The atomate unit tests have now been fixed in the master branch.

Would you be able to pull the latest changes from master and update this branch so we can check no tests are failing.

Best,
Alex

mhsiron:
> Would you be able to pull the latest changes from master and update this branch so we can check no tests are failing.

Would that be the same as just submitting a new PR with the same changes? (Still a bit new to git)

utf:
If you are in your atomate folder the commands would be:

```
git checkout patch-2
git pull upstream master
git push
```

The first command will not be necessary if you are already on the patch-2 branch but it will not hurt either way.

mhsiron:
done! thanks!

utf:
Thank you!

PR #320, opened 2019-11-20
Fix atomate tests

utf:


utf:
Atomate tests are now passing and this should be ready to merge.

I had to skip `TestAdsorptionWorkflow.test_wf_functions` as the expected output of the slab generation really isn't clear to me.

PR #336, opened 2019-12-03
Add set_queue_options powerup

rkingsbury:
## Summary

Added a powerup that allows the `walltime`, `time_min`, and `qos` queue submission options to be modified on a per-workflow basis. 

I did not make the powerup general (e.g., able to include arbitrary `SBATCH` arguments like number of nodes) because those changes would also require modification of the FWorker.

Tested and confirmed working on the Cori KNL queue. I'll note that I have only tested single-firework workflows.

### DIFF
--- a/atomate/vasp/powerups.py
+++ b/atomate/vasp/powerups.py
@@ -293,7 +293,41 @@ def clear_modify(original_wf, fw_name_constraint=None):
         original_wf.fws[idx_fw].tasks.pop(idx_t)
     return original_wf
 
+def set_queue_options(original_wf, walltime=None,time_min=None,qos=None, 
+                          fw_name_constraint=None, task_name_constraint=None):
+    """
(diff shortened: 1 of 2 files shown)


atomate/vasp/powerups.py
```diff
@@ -286,16 +352,68 @@ def clear_modify(original_wf, fw_name_constraint=None):
...
+    walltime=None,
+    time_min=None,
+    qos=None,
+    fw_name_constraint=None,
+    task_name_constraint=None,
+):
+    """
+    Modify queue submission parameters of Fireworks in a Workflow. 
+    
+    This powerup overrides paramters in the qadapter file by setting values in the 'queueadapter' key of a Firework spec. For example, the walltime
```
utf: Can you make your docstrings compatible with PEP8 (i.e., line length)

utf:
Thanks!

PR #349, opened 2019-12-17
Magnetism bug fix

mkhorton:
## Summary

A bad commit crept into master in the magnetism workflow; my apologies for that. This fixes the issue.


mkhorton:
(I think it was missed because the magnetism tests might be skipped by CI if enumlib isn't installed.)

PR #377, opened 2020-02-14
Add ModifyKpoints Firetask and associated VASP powerup

mhsiron:
## Summary
This was a useful feature for me, thought it might be useful for some other people.

Added the ModifyKpoints Firetask to modify a Kpoints object based on a dictionary that matches its attributes. Added an associated VASP powerup to do this at a workflow level.

### DIFF
--- a/atomate/vasp/powerups.py
+++ b/atomate/vasp/powerups.py
@@ -298,8 +298,8 @@ def add_modify_kpoints(
 
     Args:
         original_wf (Workflow)
-        modify_kpoints_params (dict) - dict of parameters for ModifyKpoints.
-        fw_name_constraint (str) - Only apply changes to FWs where fw_name
+        modify_kpoints_params (dict): dict of parameters for ModifyKpoints.
+        fw_name_constraint (str): Only apply changes to FWs where fw_name
         contains this substring.
 
     Returns:
--- a/atomate/vasp/powerups.py
+++ b/atomate/vasp/powerups.py
@@ -14,7 +14,8 @@ from atomate.vasp.firetasks.run_calc import (
     RunNoVasp,
 )
 from atomate.vasp.firetasks.neb_tasks import RunNEBVaspFake
-from atomate.vasp.firetasks.write_inputs import ModifyIncar, ModifyPotcar
+from atomate.vasp.firetasks.write_inputs import ModifyIncar, ModifyPotcar, \
+    ModifyKpoints
 from atomate.vasp.firetasks.parse_outputs import JsonToDb
 from atomate.vasp.config import (
     ADD_NAMEFILE,
@@ -284,6 +285,37 @@ def add_modify_incar(
     for idx_fw, idx_t in idx_list:
         original_wf.fws[idx_fw].tasks.insert(
             idx_t, ModifyIncar(**modify_incar_params)
+        )
+    return original_wf
+
+def add_modify_kpoints(
+    original_wf, modify_kpoints_params=None, fw_name_constraint=None
+):
+    """
+    Every FireWork that runs VASP has a ModifyKpoints task just beforehand. For
+    example, allows you to modify the KPOINTS based on the Worker using env_chk
+    or using hard-coded changes.
+
+    Args:
+        original_wf (Workflow)
+        modify_kpoints_params (dict) - dict of parameters for ModifyKpoints.
+        fw_name_constraint (str) - Only apply changes to FWs where fw_name
+        contains this substring.
+
+    Returns:
+       Workflow
+    """
+    modify_kpoints_params = modify_kpoints_params or {
+        "kpoints_update": ">>kpoints_update<<"
+    }
+    idx_list = get_fws_and_tasks(
+        original_wf,
+        fw_name_constraint=fw_name_constraint,
+        task_name_constraint="RunVasp",
+    )
+    for idx_fw, idx_t in idx_list:
+        original_wf.fws[idx_fw].tasks.insert(
+            idx_t, ModifyKpoints(**modify_kpoints_params)
         )
     return original_wf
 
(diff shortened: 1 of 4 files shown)


utf:
Thanks for this @mhsiron 

Please can you add unit tests for both the powerup and the FireTask. You can see examples of similar tests here:

powerup: https://github.com/hackingmaterials/atomate/blob/eedbdd74d42fbfbcc36cab2bcd05c0b80dd13098/atomate/vasp/tests/test_vasp_powerups.py#L84

FireTask: https://github.com/hackingmaterials/atomate/blob/eedbdd74d42fbfbcc36cab2bcd05c0b80dd13098/atomate/vasp/firetasks/tests/test_write_vasp.py#L119

Also, please ignore the failing py3devtests.

atomate/vasp/powerups.py
```diff
@@ -287,6 +288,37 @@ def add_modify_incar(
...
+    original_wf, modify_kpoints_params=None, fw_name_constraint=None
+):
+    """
+    Every FireWork that runs VASP has a ModifyKpoints task just beforehand. For
+    example, allows you to modify the KPOINTS based on the Worker using env_chk
+    or using hard-coded changes.
+
+    Args:
+        original_wf (Workflow)
+        modify_kpoints_params (dict) - dict of parameters for ModifyKpoints.
```
utf: Can you change the hyphens ("-") to colons, as otherwise these doc string will not be rendered properly in the documentation.

I.e.

```
modify_kpoints_params (dict): dict of parameters for ModifyKpoints.
```

mhsiron: Done! Thanks for your help!

utf:
Thanks!

mkhorton:
This seems useful -- we were having the same issue but with "modify user functional" recently with the HSE FW.

I still think there's a more elegant solution out there though, in terms of the Matryoshka nightmare of all our workflow/fireworks/firetasks kwargs. A more sane base firework for VASP calculations to enforce a uniform method signature might help, but maybe there's a better way to introspect a workflow object and change kwargs dynamically? Once a workflow object is created, we could build a capability to drill down (either by index, or by a selection rule) to individual firetasks and change their available options. If we typed the firetask spec, we could even build dynamic validation into it.

mkhorton:
(And to pre-empt a comment that this is what powerups already too -- the problem I have with powerups is that (a) a new one has to be written essentially for every kwarg, (b) they rely too heavily on regex, and it's not impossible a badly written powerup or powerup input could clobber a workflow.)

utf:
I definitely agree with the sentiment. Using a standardized VASP firework will reduce the mental complexity significantly.

To address a): We could add a general `update_firetask` powerup which is capable of updating the kwargs of any firetask. It would just be a very simple wrapper around the common powerup boilerplate.

Something like

```python
def update_firetask(
    original_wf,
    firetask_name_constraint,
    update_params,
    fw_name_constraint=None,
):
    idx_list = get_fws_and_tasks(
        original_wf,
        fw_name_constraint=fw_name_constraint,
        task_name_constraint=firetask_name_constraint,
    )
    for idx_fw, idx_t in vasp_fws_and_tasks:
        original_wf.fws[idx_fw].tasks[idx_t].update(update_params)
    
    return update_params
```

For b), do you have a preference for how to drill down into specific fireworks/tasks?

mkhorton:
Yes, I think an `update_firetasks` method would be really useful. I was thinking something like:

`workflow.update_firetasks(selection: List, name_constraint: str, key: str, value)` # updates a list of FireTasks that match the selection and constraint

Where `selection` could be a list of FireWorks or FireTasks (the classes themselves, for rigorous filtering). The name constraint which is as it is now, just a regex on FireWork name, the key is then a single kwarg name.

If we start typing the specs, we could also then do validation, which I think could be really compelling.

For usability, this could be split into two methods also, maybe a selection method that just returns a list of FireTasks within that workflow. The FireTasks themselves could have a dot syntax (eg .potcar_functional = ... etc)

Thoughts?

utf:
I can see that being able to select specific classes might be more rigorous, but in practice do the regex name constraints end up causing problems? You can always specify the full firework/firetask name if you want to be sure.

I'm not quite sure of the use cases where your proposal will give more flexibility over the powerup I suggested (which also enables you to update multiple parameters simultaneously).

As for typing the specs, I think that would be very useful, although it would require fundamentally rewriting how fireworks works (everything is no longer a dict)? I would say that this discussion is probably better suited for the Fireworks issue tracker rather than atomate.

mkhorton:
> I'm not quite sure of the use cases where your proposal will give more flexibility over the powerup I suggested (which also enables you to update multiple parameters simultaneously).

Yes, honestly I'd take any option here, I think your solution is perfectly good and very fast to implement too which is a bonus.

I think the main benefit of a more complicated solution is that we could make firetask options more discoverable -- e.g. if you get a selection first, you get a firetask object, and we could have tab-completion for what args you can change.

> As for typing the specs, I think that would be very useful, although it would require fundamentally rewriting how fireworks works (FireTasks are no longer just dicts)?

I don't think so. To take an example, take `RunVASPCustodian`, we would change:

`required_params = ["vasp_cmd"]`

to:

`required_params = {"vasp_cmd": str}`

If `required_params` is a list, Fireworks would work as it always has, if it's a dict, it'd do the typing validation also. So this would be a completely backwards-compatible change.

> I would say that this discussion is probably better suited for the Fireworks issue tracker rather than atomate.

Yes -- completely fair.

utf:
Type checking using `required_params = {"vasp_cmd": str}` is a nice idea.

> we could have tab-completion for what args you can change.

I think this would be difficult without rewriting the FireTasks to be non-dicts. You could make it so that you can access firetask properties using dot notation quite easily by defining `__getattr__`, something like:

```
    def __getattr__(self, item):
        return self[item]
```

But that wouldn't support tab completion.

PR #383, opened 2020-03-15
[WIP] Integration for CP2K

nwinner:
## Summary

Integration of CP2K into the atomate framework. This WIP PR is here so that maintainers can have a sense of what I'm proposing to add.

New firetasks, fireworks, workflows, and database/drone classes for cp2k. Meant to be used in conjunction with the new pymatgen.io.cp2k module (also a work in progress).

## TODO (if any)

More workflows/fireworks are needed. There have also been a few issues with gluing tasks with file exchange. The drone also needs to include the optional Dos, bader, and cube parsing that will be needed.


atomate/cp2k/firetasks/write_inputs.py
```diff
@@ -0,0 +1,76 @@
+# coding: utf-8
+
+from __future__ import division, print_function, unicode_literals, absolute_import
+
+"""
+This module defines tasks for writing vasp input sets for various types of vasp calculations
```
utf: Update please.

atomate/cp2k/firetasks/write_inputs.py
```diff
@@ -0,0 +1,76 @@
...
+        else:
+            cis_cls = load_class("pymatgen.io.cp2k.sets", self["cp2k_input_set"])
+            cis = cis_cls(self["structure"], **self.get("cp2k_input_params", {}))
+        cis.write_file(input_filename='cp2k.inp', output_dir='.')
+
+
+@explicit_serialize
+class WriteCp2kFromPrevious(FiretaskBase):
+
+    optional_params = ['cp2k_input_params', 'prev_calc_loc', 'original_input_filename',
```
utf: `cp2k_input_params` and `new_input_filename` are unused here.

atomate/cp2k/firetasks/glue_tasks.py
```diff
@@ -0,0 +1,147 @@
...
+
+__author__ = 'Nicholas Winner'
+__email__ = 'nwinner@berkeley.edu'
+
+
+@explicit_serialize
+class UpdateStructureFromPrevCalc(FiretaskBase):
+    """
+    Using the location of a previous calculation. The CP2K output parser will
+    get the final structure from a previous calculation and update this FW's
```
utf: Is there a reason why you are updating the structure by updating the firework spec rather than either:
1. Just writing the structure to file
2. Doing the structure updating as part of `WriteCp2kFromPrevious`.

I think option 2 makes the most sense. If you only want to copy the structure and nothing else from the previous calculation, you could add options to that effect in `WriteCp2kFromPrevious`.

atomate/cp2k/fireworks/core.py
```diff
@@ -0,0 +1,191 @@
...
+            cp2ktodb_kwargs (dict): kwargs to pass to Cp2kToDb
+            \*\*kwargs: Other kwargs that are passed to Firework.__init__.
+        """
+        t = []
+
+        # Set up the name, the input set, and how cp2k will be pushed to DB
+        fw_name = "{}-{}".format(structure.composition.reduced_formula if structure else "unknown", name)
+
+        cp2k_input_set = cp2k_input_set or StaticSet(structure, **cp2k_input_set_params)
+        cp2k_input_set.project_name = name  # Currently, this assertion is needed to keep all FWs in a workflow
```
utf: What is the problem you are trying to solve by using `project_name`?

atomate/cp2k/firetasks/glue_tasks.py
nwinner: Hey Alex, number 2 probably does make more sense. Number 1 isn't the case because cp2k doesn't use the same formatting as VASP where a separate file holds the structure. Generally cp2k has everything in a single input file. This can actually be modified, but its not a cp2k handles stuff by default so i didn't want to complicate that until a later time.

atomate/cp2k/fireworks/core.py
nwinner: The CP2K output files are named according to the "PROJECT_NAME" variable inside the input file. Notably, the wavefunction and DOS files have the project name prepended in their filename. So, in order to deal with this we would need to either (a) check for the project name every time you try to find a file (like a restart file) to copy over, (b) use a universal project name so that the files still have the same format but with a standard prefix, (c) override the file name in the cp2k input file so that it adheres to some agreed upon overall file name. I chose to go with (b) as the easiest one, but I'm open to changing it.

atomate/cp2k/database.py
```diff
@@ -0,0 +1,179 @@
...
+from atomate.utils.utils import get_logger
+
+__author__ = 'Kiran Mathew'
+__credits__ = 'Anubhav Jain'
+__email__ = 'kmathew@lbl.gov'
+
+logger = get_logger(__name__)
+
+
+class Cp2kCalcDb(CalcDb):
```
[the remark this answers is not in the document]
nwinner: Its not different at all I think, a more general VASPCalcDb would be a good idea.

atomate/cp2k/fireworks/core.py
mkhorton: Note we're going to have the same question for my (perpetual backburner project) of CASTEP support, so interested in whatever you settle on. I might pick the seed name to be the firework id, unless that has issues I'm not anticipating.

atomate/cp2k/firetasks/glue_tasks.py
utf: Ok great. Let's go with option 2!

atomate/cp2k/fireworks/core.py
utf: Ok, thanks for clarifying that. With option b) are you using the same project name for EVERY calculation or is it specific to the calculation type?  Maybe what I am describing is option c?

Are there any downsides to using a fixed project name, and just doing the calculations in different directories? That way there is never any ambiguity over which files to load. it seems like this would remove a lot of the complexity and make the parsing and file copying more similar to how it is handled in the VASP module?

nwinner: Currently option (b) uses the same name for the workflow. So you could name the files according to "Static" for static WFs as an example. Just to give you a sense of what the output files are like. If you did that you would get a directory like this (for Si):

launcher-X:
- Static-ALPHA_k1-1.pdos  # elemental dos for atom kind 1, Spin up channel   
- Static-BETA_k1-1.pdos     # elemental dos for atom kind 1, Spin down channel    
- Static-ALPHA_list1-1.pdos  # Site projected dos onto site list 1, spin up                                   
- Static-ALPHA_list2-1.pdos  # Site projected dos onto site list 2, spin up                                 
- Static-BETA_list1-1.pdos  # Site projected dos onto site list 1, spin down                      
- Static-BETA_list2-1.pdos     # Site projected dos onto site list 2, spin down                                            
- Static-RESTART.wfn     # Restart wavefunction file
- Static-RESTART.wfn.bak-1  # Backup wavefunction file (dumped intermittently)                              
- Static-v_hartree-1_0.cube.  # Hartree (electrostatic potential)
- Static-ELECTRON_DENSITY-1_0.cube   # Electron density      
- Static-SPIN_DENSITY-1_0.cube   # spin density

So you can see that it changes the naming, but its pretty repeatable. Something I've been doing in the pymatgen output parsing is globbing to find files containing, say, "pdos", to identify the pdos files and globbing for "RESTART" to get the wavefunction file name. We could think about doing this instead of trying to record what the file name is going to be.

PR #391, opened 2020-03-23
[WIP] Exchange wflow

ncfrey:
## Summary
@utf 

The exchange wflow is an extension to the magnetism wflow that enables calculation of exchange parameters and critical temperatures in magnetic materials. 

* Map pre-computed magnetic orderings to a classical Heisenberg model and solve for exchange parameters to any desired order (nearest neighbor, next-nearest neighbor, etc.).
* Estimate an effective J parameter and mean field Tc from only the lowest energy FM and AFM orderings.
* Parse outputs and interface with VAMPIRE package to do Monte Carlo simulations of Tc.

## TODO (if any)

* The static calculation step should be noncollinear, with magnetic moments constrained along some axis and with magnitudes fixed at the values from the ground state ordering. 
* More computationally demanding and accurate methods for computing exchange should be included (e.g. RPA, magnetic force theorem, etc.)


atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+            summaries.append(summary)
+
+        mmdb.collection = mmdb.db["magnetic_orderings"]
+        mmdb.collection.insert(summaries)
+
+        logger.info("Magnetic orderings calculation complete.")
+
+
+@explicit_serialize
+class HeisenbergModelMapping(FiretaskBase):
```
utf: So this class seems to do two things at once. It computes the the exchange parameters AND adds the model to the database.

If the mapping is trivial then I think it would be best to just treat this is a ToDb FireTask (and rename it accordingly). If the mapping can take a considerable amount of time, say > 10 mins then maybe this could be broken up into two tasks. One to do the mapping and writing the mapping to a file, and other to add the data to the database.

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+
+    """
+
+    required_params = ["db_file", "exchange_wf_uuid", "parent_structure", "avg"]
+    optional_params = []
+
+    def run_task(self, fw_spec):
+
+        db_file = env_chk(self["db_file"], fw_spec)
+        wf_uuid = self["exchange_wf_uuid"]
```
utf: In keeping with the magnetism workflow, can you rename this parameter in all of your firetasks to just wf_uuid...

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+        parent_structure (Structure): Structure object.
+        avg (bool): <J> exchange param only.
+
+    TODO:
+        * More robust convergence check
+
+    """
+
+    required_params = ["db_file", "exchange_wf_uuid", "parent_structure", "avg"]
+    optional_params = []
```
utf: You can remove this line.

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+class VampireMC(FiretaskBase):
+    """
+    If there is a suitable Heisenberg model, run Monte Carlo to compute the critical temperature.
+
+    Args:
+        db_file (str): path to file containing the database credentials.
+        exchange_wf_uuid (int): Unique id for record keeping.
+        parent_structure (Structure): Structure object with magmoms.
+        mc_settings (dict): A configuration dict for monte carlo.
+        avg (bool): Only <J> exchange param.
```
utf: Rename this to `average`

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+                hmodel = None
+
+        # Update FW spec with converged hmodel or None
+        update_spec = {"converged_heisenberg_model": hmodel}
+
+        return FWAction(update_spec=update_spec)
+
+
+@explicit_serialize
+class VampireMC(FiretaskBase):
```
utf: Again, if this is likely to take some time, can we break up the running and ToDb steps.

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+        if fw_spec.get("tags", None):
+            task_doc["tags"] = fw_spec["tags"]
+
+        # Vampire monte carlo settings
+        mc_box_size = mc_settings["mc_box_size"]
+        equil_timesteps = mc_settings["equil_timesteps"]
+        mc_timesteps = mc_settings["mc_timesteps"]
+
+        # Get a converged Heisenberg model if one was found
+        # if fw_spec["converged_heisenberg_model"]:
```
utf: Is there a reason why you don't use this approach?

atomate/vasp/fireworks/exchange.py
```diff
@@ -0,0 +1,157 @@
...
+    ):
+        """
+        Takes a set of low-energy magnetic orderings and energies and maps them to a Heisenberg Model to compute exchange params.
+
+        Args:
+            exchange_wf_uuid (int): Unique id for record keeping.
+            parent_structure (Structure): Magnetic ground state.
+            parents (FireWorks): Parent FWs.
+            db_file (str): Path to file containing db credentials.
+            heisenberg_settings (dict): A config dict for Heisenberg model mapping.
```
utf: Can you add more detail here with the supported options.

atomate/vasp/fireworks/exchange.py
```diff
@@ -0,0 +1,157 @@
...
+class VampireCallerFW(Firework):
+    def __init__(
+        self,
+        exchange_wf_uuid,
+        parent_structure,
+        parents,
+        db_file=DB_FILE,
+        mc_settings=None,
+        name="vampire caller",
+        c=None,
```
utf: Again, remove.

atomate/vasp/fireworks/exchange.py
```diff
@@ -0,0 +1,157 @@
...
+        avg=True,
```
utf: Please rename to average throughout. We prefer explicit variable names where possible.

atomate/vasp/workflows/base/exchange.py
```diff
@@ -0,0 +1,353 @@
...
+        db_file=DB_FILE,
+        name="Exchange WF",
+    ):
+        """Workflow for computing exchange parameters.
+
+        This workflow takes a set of magnetic orderings and their energies from MagneticOrderingsWF and runs static calculations of magnetic orderings in the ground state geometry. The static orderings/energies are then fit to a classical Heisenberg Hamiltonian to compute exchange parameters. The critical temperature can then be calculated with Monte Carlo.
+
+        Optionally, only the lowest energy FM and AFM configurations can be used to compute the average exchange interaction, <J>, without any static calculations.
+
+        Args:
```
utf: Please add all args and kwargs to the docstring.

atomate/vasp/workflows/base/exchange.py
```diff
@@ -0,0 +1,353 @@
...
+                        name=name + "static",
+                    )
+                )
+
+                analysis_parents.append(static_fws[-1])
+
+            # Magnetic ordering analysis that generates 'magnetic_orderings'
+            # collection
+            fw_analysis = Firework(
+                ExchangeToDB(
```
utf: What is the reason for using this versus MagneticOrderingsToDB?

atomate/vasp/workflows/base/exchange.py
```diff
@@ -0,0 +1,353 @@
...
+__email__ = "ncfrey@lbl.gov"
+__status__ = "Development"
+__date__ = "July 2019"
+__version__ = 0.1
+
+
+class ExchangeWF:
+    def __init__(
+        self,
+        magnetic_structures,
```
utf: Can you add a check in the init that all structures have magnetism information and if not raise an error.

atomate/vasp/workflows/base/exchange.py
```diff
@@ -0,0 +1,353 @@
...
+
+__author__ = "Nathan C. Frey"
+__maintainer__ = "Nathan C. Frey"
+__email__ = "ncfrey@lbl.gov"
+__status__ = "Development"
+__date__ = "July 2019"
+__version__ = 0.1
+
+
+class ExchangeWF:
```
utf: It seems like this overlaps considerably with `MagneticOrderingsWF`. Could this be simplified by adding an option to only perform static calculations in `MagneticOrderingsWF` and then use that internally to generate the most of the fireworks when `static=True`.

@mkhorton  I don't know if you have any thoughts on this?

ncfrey:
> Hi @ncfrey. Thanks for submitting this pull request. It looks in pretty good shape!
> 
> I've made some initial comments and suggestions. Please let me know if any of my comments aren't clear.

Awesome, thanks very much, Alex! I will get started on addressing these comments.

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+
+
+@explicit_serialize
+class ExchangeToDB(FiretaskBase):
+    """
+    Used to aggregate tasks docs from magnetic ordering workflow.
+
+    Args:
+        db_file (str): path to the db file that holds your tasks
+        collection and that you want to hold the magnetic_orderings
```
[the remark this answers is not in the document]
ncfrey: Sorry about that, fixed now!

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+        used to make it easier to retrieve task docs
+        parent_structure: Structure of parent crystal (not magnetically
+        ordered)
+    """
+
+    required_params = [
+        "db_file",
+        "wf_uuid",
+        "parent_structure",
+        "perform_bader",
```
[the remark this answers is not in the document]
ncfrey: Added docstrings

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+        "scan",
+    ]
+    optional_params = ["origins", "input_index"]
+
+    def run_task(self, fw_spec):
+
+        uuid = self["wf_uuid"]
+        db_file = env_chk(self.get("db_file"), fw_spec)
+        to_db = self.get("to_db", True)
```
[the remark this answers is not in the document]
ncfrey: Yes, fixed

atomate/vasp/firetasks/exchange_tasks.py
```diff
@@ -0,0 +1,458 @@
...
+from pymatgen.command_line.vampire_caller import VampireCaller
+
+__author__ = "Nathan C. Frey"
+__email__ = "ncfrey@lbl.gov"
+
+logger = get_logger(__name__)
+
+
+@explicit_serialize
+class ExchangeToDB(FiretaskBase):
```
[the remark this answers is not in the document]
ncfrey: Yes haha, I think the original plan was to add some additional info but that wasn't necessary and I never changed it. I'll get rid of it and switch to `MagneticOrderingsToDB`.

atomate/vasp/firetasks/exchange_tasks.py
ncfrey: The timing for the mapping is highly variable so I'll split it into two tasks.

atomate/vasp/firetasks/exchange_tasks.py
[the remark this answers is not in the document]
utf: Ok! In that case, I think it would be useful to have the ability to distinguish magnetic orderings calculated with relaxations vs at a fixed geometry. Could you:

1. Implement the formatting and docstring changes I mentioned above to the `MagneticOrderingsToDB` firetask.
2. Add an option to `MagneticOrderingsToDb` called `additional_fields` which just gets added to the task doc. This is a common pattern in atomate, for example you can copy the implementation of `BoltztrapToDb`.
3. Make your workflow submit additional data indicating that the structure was not relaxed before hand. How you handle I don't have any opinions about.

That way we can easily filter out magnetic structures from the exchange workflow vs magnetism workflow.

@mkhorton Does that sound reasonable to you?

atomate/vasp/firetasks/exchange_tasks.py
utf: Ok great. Similar to the `BoltztrapToDb` firetask, you can just presume you are in the correct folder with the mapping written to a json file.

atomate/vasp/firetasks/exchange_tasks.py
ncfrey: Yes, I'll split them up.

atomate/vasp/firetasks/exchange_tasks.py
ncfrey: At the moment, because of the issues with the input static calculations, this sort of convergence testing is really not very informative or useful. I think with constrained magnetic moments it will be worth doing.

atomate/vasp/fireworks/exchange.py
ncfrey: Done!

atomate/vasp/workflows/base/exchange.py
ncfrey: Yes, major overlap. The `MagneticOrderingsWF` could be modified to offer a static-only option, and also a "static-only with constrained noncollinear moments" option.

utf: Ok. If it's not too much trouble to at least add the static only option I think that would help reduce the complexity of this pull request a lot.

atomate/vasp/firetasks/exchange_tasks.py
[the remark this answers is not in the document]
ncfrey: I will go ahead and make these changes, it shouldn't be difficult and it will make it easier in the future to modify the static input step.

@utf and @mkhorton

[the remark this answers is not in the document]
mkhorton: I'm on board with this, have been chatting with @ncfrey about our options offline, `MagneticOrderingsToDb` will need to be updated a little but otherwise it should be fine

atomate/vasp/firetasks/exchange_tasks.py
ncfrey: Done! 👍

atomate/vasp/workflows/base/exchange.py
utf: Hi @ncfrey. I'm a bit confused about what's happened to this workflow, it seems like you remove the static option?

ncfrey: Hi @utf, after discussing with @mkhorton it seemed like it would be easiest to kind of partition all magnetic ordering calculations into the `MagneticOrderingsWF`, so I moved the static only option there. 

We are going to discuss how that can be extended/modified to include the more robust options I've mentioned.

My idea was that now the `ExchangeWF` is totally agnostic to how the magnetic structures and energies are generated, which was a goal I had at the beginning. Let me know if anything is unclear.

utf: That makes sense! Thanks for clearing that up.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,352 @@
...
+        tol (float): distance tolerance for similar bonds.
+        average (bool): <J> only or full NN, NNN HeisenbergModel.
+
+    """
+
+    required_params = [
+        "db_file",
+        "wf_uuid",
+        "parent_structure",
+        "cutoff",
```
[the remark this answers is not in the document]
ncfrey: Yeah, I can grab those from the model instance 👍 I'll change that.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,352 @@
...
+    ]
+
+    # If only doing <J>, give the original structure/energy inputs
+    optional_params = ["structures", "energies"]
+
+    def run_task(self, fw_spec):
+
+        db_file = env_chk(self["db_file"], fw_spec)
+        wf_uuid = self["wf_uuid"]
+        formula = self["parent_structure"].formula
```
utf: It seems formula and formula_pretty are never used in this Firetask and parent structure is only required to get the formulas.

Can you remove these lines and also remove parent_structure as a required param.

atomate/vasp/firetasks/exchange.py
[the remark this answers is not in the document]
utf: Ok great.  It also seems like the average parameter isn't required by the firetask any more.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,352 @@
...
+
+    If not, it doesn't make sense to do Monte Carlo and the system is either
+    1) not well described by the Heisenberg Model (perhaps the magnetic
+    moments are not localized) or 2) not correctly captured by the provided
+    low-energy magnetic orderings.
+
+    Args:
+        db_file (str): path to file containing the database credentials.
+        wf_uuid (int): Unique id for record keeping.
+        parent_structure (Structure): Structure object.
```
utf: Can you double check whether parent_structure is actually required. Again, it seems like it is only used for formula_pretty which itself isn't used anywhere else.

If you do need the parent structure, can you get it from the database without requiring it as a parameter.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,352 @@
...
+@explicit_serialize
+class VampireMC(FiretaskBase):
+    """
+    If there is a suitable Heisenberg model, run Monte Carlo to compute the
+    critical temperature.
+
+    Args:
+        db_file (str): path to file containing the database credentials.
+        wf_uuid (int): Unique id for record keeping.
+        parent_structure (Structure): Structure object with magmoms.
```
utf: Similar comment on parent_structure here.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,352 @@
...
+        "db_file",
+        "wf_uuid",
+        "parent_structure",
+    ]
+
+    def run_task(self, fw_spec):
+
+        db_file = env_chk(self["db_file"], fw_spec)
+        wf_uuid = self["wf_uuid"]
+        formula_pretty = self["parent_structure"].composition.reduced_formula
```
utf: And here... Can this not just be taken from the database.

atomate/vasp/firetasks/exchange.py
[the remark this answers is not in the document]
ncfrey: I cleaned that up but kept "cutoff" since it's used to distinguish between different .json models.

atomate/vasp/firetasks/exchange.py
ncfrey: Yep same thing, the model is being grabbed from the db and parent structure isn't needed. I'll take that out where it isn't needed.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,336 @@
...
+@explicit_serialize
+class VampireMC(FiretaskBase):
+    """
+    If there is a suitable Heisenberg model, run Monte Carlo to compute the
+    critical temperature.
+
+    Args:
+        db_file (str): path to file containing the database credentials.
+        wf_uuid (int): Unique id for record keeping.
+        mc_settings (dict): A configuration dict for monte carlo.
```
utf: Can you point to where the users can find the valid options. I.e., say something like. See pymatgen.x.y.VampireCaller for supported options.

utf: Also can this be an optional argument? See comment below.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,336 @@
...
+        mc_box_size = mc_settings["mc_box_size"]
+        equil_timesteps = mc_settings["equil_timesteps"]
+        mc_timesteps = mc_settings["mc_timesteps"]
+
+        # Get a converged Heisenberg model if one was found
+        # if fw_spec["converged_heisenberg_model"]:
+        #     hmodel = HeisenbergModel.from_dict(fw_spec["converged_heisenberg_model"])
+
+        vc = VampireCaller(
+            mc_box_size=mc_box_size,
```
utf: Rather than expanding the mc_settings above, can you just do:

```python
vc = VampireCaller(
    hm=hmodel,
    avg=average,
    **mc_settings
)
```

That way if a user doesn't include one of the options, the VampireCaller will fall through back on the default settings.

Secondly, if you make `mc_settings` optional with a default value of an empty dict (`mc_settings = self.get("mc_settings, {})`)  this will again nicely support falling back on the default arguments.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,336 @@
...
+    ]
+
+    def run_task(self, fw_spec):
+
+        db_file = env_chk(self["db_file"], fw_spec)
+        wf_uuid = self["wf_uuid"]
+
+        name = "heisenberg_model_" + str(self["cutoff"])
+        hmodel = loadfn(name + ".json")
+        #hmodel = HeisenbergModel.from_dict(hmodel_dict)
```
utf: The loadfn should be the best way to do this. You can probably remove this commented line.

atomate/vasp/fireworks/exchange.py
```diff
@@ -0,0 +1,181 @@
...
+            mc_box_size (float): MC simulation box size in nm.
+            equil_timesteps (int): Number of MC equilibration moves.
+            mc_timesteps (int): Number of MC moves for averaging.
+
+        Args:
+            wf_uuid (int): Unique id for record keeping.
+            parent_structure (Structure): Magnetic ground state.
+            parents (FireWorks): Parent FWs.
+            db_file (str): Path to file containing db credentials.
+            mc_settings (dict): A configuration dict for monte carlo.
```
utf: I guess the type hinting for this should be `Optional[dict]`. And again, having this as option fits nicely in with the above!

atomate/vasp/workflows/base/exchange.py
```diff
@@ -0,0 +1,183 @@
...
+            parent_structure=self.structures[0],
+            db_file=c["DB_FILE"],
+            heisenberg_settings=c["heisenberg_settings"],
+            parents=None,
+            structures=self.structures,
+            energies=self.energies,
+        )
+
+        # Vampire Monte Carlo
+        mc_settings = {
```
utf: These are the same as the `VampireCaller` defaults. Currently this results in duplication of settings. Ideally if you change the default VampireCaller defaults to make them better, you'd then have to change them in atomate too. If you follow the above by making this an optional argument and then don't set it, we can use the Pymatgen defaults for free.

ncfrey:
> Hi @ncfrey
> 
> This is looking great and I think it's almost there.
> 
> What are your thoughts on adding a present workflow that automatically does the magnetism workflow and then feeds the results into your exchange workflow?

That sounds good to me. I think it might make sense to offer the option in the magnetism wflow to run exchange immediately as a follow on.

@mkhorton

utf:
> That sounds good to me. I think it might make sense to offer the option in the magnetism wflow to run exchange immediately as a follow on.

That makes sense as the exchange workflow stands. I'm worried once we start adding additional options to do constrained magnetic moments etc, the magnetism workflow might start getting a bit overloaded.

The other issue is that it somewhat goes against one of the principles of atomate, which is that the "base" workflows should be the bare minimum set of calculations needed to get a property and the "preset" workflows are fully featured workflows.

This distinction is a little fuzzy already with the magnetism workflow (there is no preset version), so I'm open to suggestions here.

ncfrey:
> > That sounds good to me. I think it might make sense to offer the option in the magnetism wflow to run exchange immediately as a follow on.
> 
> That makes sense as the exchange workflow stands. I'm worried once we start adding additional options to do constrained magnetic moments etc, the magnetism workflow might start getting a bit overloaded.
> 
> The other issue is that it somewhat goes against one of the principles of atomate, which is that the "base" workflows should be the bare minimum set of calculations needed to get a property and the "preset" workflows are fully featured workflows.
> 
> This distinction is a little fuzzy already with the magnetism workflow (there is no preset version), so I'm open to suggestions here.

Oh ok, yeah I think I see what you mean here. I'm going to talk to @mkhorton on Monday and we can discuss this. I think it would be a good idea in general to have some preset magnetism wflows anyway.

atomate/vasp/workflows/base/exchange.py
ncfrey: Sounds good, I'll do the same with the `HeisenbergMapper` defaults as those are defined in pymatgen as well.

ncfrey:
Hey @utf, after discussing with @mkhorton what we're planning to do is keep this wflow and the magnetism wflow as they are, and then Guy Moore will build a preset wflow that combines them with the appropriate settings and the constrained calculations. How does that sound?

utf:
That sounds good to me. If you wouldn't mind implementing the last few comments I had about using the pymatgen defaults, I think this should be ready to merge

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,313 @@
...
+        energies (list): Energies / atom (eV).
+
+    Optional parameters:
+        heisenberg_settings (dict): A config dict for Heisenberg model 
+            mapping, detailed above.
+
+    """
+
+    required_params = [
+        "db_file",
```
utf: It looks like `db_file` and `wf_uuid` are no longer required btw.

ncfrey: Fixed!

ncfrey:
Hey @utf, not sure if you saw the latest commit - I took care of removing unnecessary parameters and defaulting to pymatgen defaults. Let me know if there are any more changes I should make.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
+-   repo: https://github.com/ambv/black
```
utf, line 2: Can you update this: https://github.com/psf/black

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,309 @@
...
+
+        # Get documents
+        docs = list(
+            mmdb.collection.find(
+                {"wf_meta.wf_uuid": wf_uuid}, ["heisenberg_model", "nn_cutoff"]
+            )
+        )
+
+        hmodels = [HeisenbergModel.from_dict(d["heisenberg_model"]) for d in docs]
+        hmodel = hmodels[0]  # Take the model with smallest NN cutoff
```
utf, line 256: Is this guaranteed to be the model with the smallest cutoff? Should you sort the list on cutoff size first?

atomate/vasp/workflows/tests/test_exchange_workflow.py
```diff
@@ -0,0 +1,64 @@
...
+        c["mc_settings"] = self.mc_settings
+        c["DB_FILE"] = self.db_file
+
+        wf = ExchangeWF(
+            magnetic_structures=self.structures,
+            energies=self.energies,
+            db_file=self.db_file,
+        ).get_wf(c=c)
+
+        self.assertEqual(wf.name, "Mn3Al - Exchange")
```
utf, line 60: Can you expand this into proper workflow testing. Currently the workflow is not run to check whether all fireworks/firetasks execute correctly. Once the workflow has finished you should examine the database at the end of the workflow to check that:

i) The workflow status is COMPLETED and not FIZZLED.
ii) The magnetism/exchange collections have the correct data in them.

See TestFerroelectricWorkflow for an example of how to do this (a lot of the complexity in that workflow is handling `RunFakeVasp` which you will not have in your workflow). Similar to your other tests, you can use `@requires` to avoid running the tests if vampire is not installed.

It would be good to get vampire installed on the testing environment. Is it available on conda?

.pre-commit-config.yaml
utf: Also, can you run all your new files through black once you're finished with the rest of the comments.

ncfrey: No problem! I will get started on finishing up the wflow test.

atomate/vasp/workflows/tests/test_exchange_workflow.py
ncfrey: Unfortunately, I don't think it's available through conda. The build procedure is pretty simple though, and the C++ code is here: https://github.com/richard-evans/vampire

ncfrey:
@utf I added a real wflow test, following the format in TestFerroelectricWorkflow, that checks for correct data in the exchange collection and verifies that all the FWs are COMPLETED. I'm not sure if there's an easy way to add vampire to the testing environment, but let me know if there's anything I can do to help or any changes I should make to the tests.

utf:
Thanks for adding the test so quickly. It should be very easy to get vampire installed on circleCI. Can you add below this line https://github.com/hackingmaterials/atomate/blob/c00845908729740b5b55af9b54de94097b5899a6/.circleci/config.yml#L21

and below this line: 
https://github.com/hackingmaterials/atomate/blob/c00845908729740b5b55af9b54de94097b5899a6/.circleci/config.yml#L51

These lines:
```
wget https://vampire.york.ac.uk/resources/release-5/vampire-5.0-linux.tar.gz
tar -zxvf vampire-5.0-linux.tar.gz; cd linux; ./install.sh; cd ..
```

And then hopefully vampire tests will now run.

.circleci/config.yml
```diff
@@ -19,6 +19,9 @@ jobs:
             # conda install --quiet --yes -c clinicalgraphics vtk
             conda install --quiet --yes -c conda-forge python-igraph
             conda update --quiet --all
+            wget https://vampire.york.ac.uk/resources/release-5/vampire-5.0-linux.tar.gz
+            tar -zxvf vampire-5.0-linux.tar.gz
+            export PATH=/root/linux/:$PATH
```
utf, line 24: Sorry about this, this line should be:

```
export PATH=/root/atomate/linux/:$PATH
```

.circleci/config.yml
```diff
@@ -49,6 +52,9 @@ jobs:
...
```
utf, line 57: And so should this line:

```
export PATH=/root/atomate/linux/:$PATH
```

utf:
So the vampire tests now run!

They are flagging an error here:

```
self = <atomate.vasp.workflows.tests.test_exchange_workflow.TestExchangeWF testMethod=test_workflow>

    @unittest.skipIf(not vampire_present, "vampire not present")
    def test_workflow(self):
        wf = self._get_simulated_wflow()
    
        self.assertEqual(wf.name, "Mn3Al - Exchange")
    
        # Heisenberg + VampireMC = 2
        self.assertEqual(len(wf.fws), 2)
    
        fw_ids = self.lp.add_wf(wf)
        rapidfire(self.lp, fworker=FWorker(env={"db_file": self.db_file}))
    
        # Exchange collection
        # mmdb = VaspCalcDb.from_db_file(self.db_file, admin=True)
        # mmdb.collection = mmdb.db["exchange"]
        exchange_collection = self.get_task_database().exchange
    
        # Check Heisenberg model mapping
        d = exchange_collection.find_one({"task_name": "heisenberg model"})
>       self._check_run(d, "heisenberg model")

../../vasp/workflows/tests/test_exchange_workflow.py:78: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <atomate.vasp.workflows.tests.test_exchange_workflow.TestExchangeWF testMethod=test_workflow>
d = None, task_name = 'heisenberg model'

    def _check_run(self, d, task_name):
        if task_name == "heisenberg model":
>           self.assertEqual(d["nn_cutoff"], 3.0)
E           TypeError: 'NoneType' object is not subscriptable

../../vasp/workflows/tests/test_exchange_workflow.py:93: TypeError
```

ncfrey:
Good news that they are running! Hm, I'm not seeing that error when I run the tests locally. It looks like it's failing to grab the task doc from the `exchange` collection. Is there a reason why
`exchange_collection = self.get_task_database().exchange` would work locally but not in CI?

utf:
I don't think that is the issue.

If you check the test logs it has this:

```
020-04-08 16:08:43,032 INFO Performing db tune-up
2020-04-08 16:08:43,105 INFO LaunchPad was RESET.
2020-04-08 16:08:43,133 INFO Added a workflow. id_map: {-8408: 1, -8407: 2}
2020-04-08 16:08:43,141 INFO Created new dir /root/atomate/atomate/utils/scratch/launcher_2020-04-08-16-08-43-141564
2020-04-08 16:08:43,141 INFO Launching Rocket
2020-04-08 16:08:43,196 INFO RUNNING fw_id: 2 in directory: /root/atomate/atomate/utils/scratch/launcher_2020-04-08-16-08-43-141564
2020-04-08 16:08:43,224 INFO Task started: {{atomate.vasp.firetasks.exchange.HeisenbergModelMapping}}.
2020-04-08 16:08:43,687 INFO Task completed: {{atomate.vasp.firetasks.exchange.HeisenbergModelMapping}} 
2020-04-08 16:08:43,767 INFO Rocket finished
----------------------------- Captured stderr call -----------------------------
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/fireworks/core/rocket.py", line 245, in run
    Rocket.update_checkpoint(lp, launch_dir, launch_id, checkpoint)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/fireworks/core/rocket.py", line 435, in update_checkpoint
    launchpad.ping_launch(launch_id, checkpoint=checkpoint)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/fireworks/core/launchpad.py", line 1525, in ping_launch
    m_launch.trackers]}})
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/collection.py", line 1003, in update_one
    session=session),
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/collection.py", line 856, in _update_retryable
    _update, session)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/mongo_client.py", line 1491, in _retryable_write
    return self._retry_with_session(retryable, func, s, None)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/mongo_client.py", line 1384, in _retry_with_session
    return func(session, sock_info, retryable)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/collection.py", line 852, in _update
    retryable_write=retryable_write)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/collection.py", line 823, in _update
    _check_write_command_response(result)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/helpers.py", line 221, in _check_write_command_response
    _raise_last_write_error(write_errors)
  File "/root/miniconda3/envs/test_env/lib/python3.7/site-packages/pymongo/helpers.py", line 203, in _raise_last_write_error
    raise WriteError(error.get("errmsg"), error.get("code"), error)
pymongo.errors.WriteError: The dotted field 'heisenberg_model_3.0' in 'state_history.0.checkpoint._all_update_spec.heisenberg_model_3.0' is not valid for storage.
```

So it seems like the string "heisenberg_model_3.0" is somewhere in firetask and is causing confusion.

atomate/vasp/firetasks/exchange.py
```diff
@@ -0,0 +1,289 @@
...
+
+        # Map system to a Heisenberg Model
+        hmapper = HeisenbergMapper(structures, energies, **heisenberg_settings)
+
+        # Get MSONable Heisenberg Model
+        hmodel = hmapper.get_heisenberg_model()
+        hmodel_dict = hmodel.as_dict()
+
+        # Update FW spec with model
+        name = "heisenberg_model_" + str(hmodel.cutoff)
```
utf, line 65: This is the line that is causing confusion.

Can you edit this line to be
```python
name = "heisenberg_model_" + str(hmodel.cutoff).replace(".", "_")
```

ncfrey: Got it, thanks!

utf:
Thanks for all your work on this pull request.  Everything looks good now, so I'll merge.

PR #405, opened 2020-05-20
Linear Response Hubbard U Workflow

guymoore13:
## Summary

This is a "work in progress" pull request for linear response LDA+U VASP workflow for calculating Hubbard U values based on the work of Cococcioni et. al.

* Addition of get_linear_response_u function to obtain linear response workflow
* Update of LinearResponseUSet & LinearResponseUFW based on Martin Siron's ( @mhsiron ) original additions
* Addition of LinearResponseUToDb firework
* Temporary PoscarPerturb class inherited from Poscar to allow individual sites to be designated as separate species
* Linear response at multiple sites
* Spin-polarized response (+ Hund J)


guymoore13:
Hi Matt and Jason (@mkhorton & @munrojm),

It's still a work in progress, but here is the initial pull request for the linear response Hubbard U workflow.

I look forward to receiving your feedback!

Thank you,
Guy

mkhorton:
Thanks @guymoore13 ! We'll take a look :)

mkhorton:
Hi Guy, haven't had time to do a full review yet, but wondering why it's necessary to duplicate the `Poscar` init? likewise a lot of duplicated code seems to be in the LinearResponseUSet `incar` property.

mkhorton:
> I will think about a way around using the regex statements. Do you think that it would be more robust to query using the distinguishing VASP parameters themselves (i.e. ICHARG, ISPIN, etc.)?

Yes, though to be fair I've used the regex method myself.

guymoore13:
I've removed the regex query and provided an option for non-spin polarized calculations. It needs a bit more testing but once that's verified it should be ready to go.

mkhorton:
Fantastic, thanks @guymoore13 !

guymoore13:
I'll use this testing as an opportunity to cleanup the previous runs and organize the analysis FWs within the database!

mkhorton:
You can also change the number of decimal places with `assertAlmostEqual`, a value of 4 or 5 decimal places should be more than sufficient and avoid machine precision issues.

guymoore13:
Thank you @mkhorton, it looks like the check finally passed for `py3test`. But maybe I should lower the precision anyways to avoid these issues for future edits(?) The recent change is that I updated the test function to check the U (&J) values computed using the different matrix inversion methods. I think that it is ready to be merged, despite the `py3devtest` fail that I may have mentioned to you. However, it appears to be an "KeyError: 'ENAUG'" issue. It appears to be elsewhere in the code, but I can't tell if this is because of something I changed.

mkhorton:
Yes, I think that's related to some changes from Ryan's SCAN workflow on the pymatgen side. I think it might have been fixed with the latest version of pymatgen, but in any case it's unrelated. Unfortunately, sometimes changes in pymatgen can cause resulting test failures in atomate which it did in this case.

mkhorton:
And agreed, lowering the precision sounds sensible regardless, though glad they're passing!

guymoore13:
Oh okay thank you. That's unfortunate, but good to know. 

Also, I should mention that the `sample_trials.json` file that I used for the analysis test is quite large (over 40 MB). I tried to whittle out the large components, such as the serialized LOCPOT, but it still seems comparatively large.

guymoore13:
I can cut out the parts of the OUTCAR that I'm not using, and the parent (non spin polarized) tasks that aren't used in the analysis.

mkhorton:
I'm not seeing `sample_tasks.json` just `sample_tasks_multisite_spinpolarized.json`, but have you tried using `dumpfn` with `.gz` at the end of your filename? It should reduce file size ~ 10x, would be easier than manually trimming the OUTCAR which will be a pain if you ever need to re-generate the test files.

guymoore13:
Yes, sorry I meant `sample_tasks_multisite_spinpolarized.json`. Oh that's very good to know, thank you for the tip. I just used `json.dump`, but I will try again with the GNU zip extension.

guymoore13:
I believe that this PR is ready to merge if you think that if you think that the code meets the appropriate standards

mkhorton:
Thanks @guymoore13! I'll have another look through (and I'll test the latest version) and then maybe, @utf would you be willing to do a final review before merging?

Would it be possible to paste a sample output document into this issue too? I think it may be helpful for discussion.

guymoore13:
Thank you Matt for your quick reply. What do you mean by "output document"? Would this be a sample output of the analysis firework?

mkhorton:
Yes, to be more precise a sample document from `HubbardHundLinRespToDb`

mkhorton:
e.g. the one generated by the current test

guymoore13:
Thanks, and sure thing. I've attached the document in `json` format (as a .txt) for the output of that example (LiNiPO4). I've included the full document, but please let me know if you would like it in a more readable format.
[sample_tasks_multisite_spinpolarized_OUTPUT.txt](https://github.com/hackingmaterials/atomate/files/5234107/sample_tasks_multisite_spinpolarized_OUTPUT.txt)

atomate/vasp/workflows/base/hubbard_hund_linresp.py
```diff
@@ -0,0 +1,639 @@
...
+__status__ = "Production"
+__date__ = "February 2020"
+
+__hubbard_hund_linresp_wf_version__ = 0.0
+
+module_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)))
+
+
+def get_wf_hubbard_hund_linresp(
+        structure, uis,
```
mkhorton, line 50: `uis` could be spelled out, should also be optional (default provided in body of `__init__`)

atomate/vasp/workflows/base/hubbard_hund_linresp.py
```diff
@@ -0,0 +1,639 @@
...
+        relax_nonmagnetic=True,
+        spin_polarized=True,
+        applied_potential_range=[-0.2, 0.2],
```
mkhorton, line 53: Mutable default argument is dangerous, should be `(-0.2, 0.2)`

atomate/vasp/workflows/base/hubbard_hund_linresp.py
```diff
@@ -0,0 +1,639 @@
...
+        c=None):
+    """
+    Compute Hubbard U (and Hund J) on-site interaction values using GGA+U
+    linear response method proposed by Cococcioni et. al.
+    (DOI: 10.1103/PhysRevB.71.035105)
+    and the spin-polarized response formalism developed by Linscott et. al.
+    (DOI: 10.1103/PhysRevB.98.235157)
+
+    Args:
+        structure:
```
mkhorton, line 68: Some arguments not documented, or inconsistent with method signature

atomate/vasp/workflows/base/hubbard_hund_linresp.py
```diff
@@ -0,0 +1,639 @@
...
+        spin_polarized: Perform spin-dependent perturbations
+        applied_potential_range: Bounds of applied potential
+        num_evals: Number of perturbation evalutaions
+        uis: user INCAR settings (must set LMAXMIX, i.e. uis = {"LMAXMIX": 4})
+        site_indices_u: List of site indices within
+    Structure indicating perturbation sites
+        species_u: List of names of species (string)
```
mkhorton, line 75: `species_u` could be elevated, and made more explicit that user is expected to provide `species_u` or `site_indices_u`

atomate/vasp/workflows/base/hubbard_hund_linresp.py
```diff
@@ -0,0 +1,639 @@
...
+    of sites to perturb; First site of that species
+    is selected in the structure
+        use_default_uvals: Use the default U values
```
mkhorton, line 78: `use_default_uvals` could perhaps be more general, e.g. so user can provide specific U values to use for other sites (this would just get passed into `user_incar_settings` ultimately, so perhaps this isn't necessary if the user can set the default uvals via `uis`)

atomate/vasp/workflows/base/hubbard_hund_linresp.py
```diff
@@ -0,0 +1,639 @@
...
+    for non-perturbed sites
+        parallel_scheme: 0 - (default) self-consistent (SCF)
```
mkhorton, line 80: More description of `parallel_scheme` would be useful, e.g. why should someone use `0` or `1`

mkhorton:
@utf do you have any comments on this PR, especially anything that would be a roadblock to getting it merged?

atomate/vasp/workflows/base/hubbard_hund_linresp.py
guymoore13: I left `uis` as a required input due to the fact that `LMAXMIX` must be specified by the user. However, should I add some simple logic to set LMAXMIX=4 if a d-block, LMAXMIX=6 if a f-block, element is present in the input `Structure`?

guymoore13: I've added a simple statement to `__init__` to set `LMAXMIX`, but please let me know if you have a better solution

mkhorton: Setting automatically based on d/f seems ideal

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -775,6 +776,590 @@ def run_task(self, fw_spec):
...
+            struct = Structure.from_dict(
+                d["calcs_reversed"][-1]["output"]['structure'])
+
+            incar_dict = d['calcs_reversed'][0]['input']['incar']
+            outcar_dict = d['calcs_reversed'][0]['output']['outcar']
+
+            # Check if task is used in LR analysis
+            use_calc = False
+            rkey = ""
+            if int(incar_dict["ICHARG"]) == 11:
```
mkhorton, line 857: This line caused a KeyError on testing. Using `.get()` allows for the logic to work even if the key is not specified.

Also, a more compact way of writing lines 857:866 might be:

```
use_calc = (int(incar_dict.get("ICHARG") == 11) or (relax_nonmagnetic and int(incar_dict.get("ISPIN")==2)
```

(This isn't more or less right, just personal preference)

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -775,6 +776,590 @@ def run_task(self, fw_spec):
...
+        keys = ['Ground state', 'NSCF', 'SCF']
+        response_dict = {'Ground state': {}, 'NSCF': {}, 'SCF': {}}
+        perturb_dict = {}
+
+        magnet_order_gs = None
+
+        for key in keys:
+
+            for i in range(num_perturb_sites):
+                response_dict[key].update({'Vup_site'+str(i): [],
```
mkhorton, line 823: There might be a better way of structuring this. In general, encoding multiple bits of information in a dict key is not ideal, since it then has to be re-parsed later to get information out (eg the perturbed site index here).

An alternative would be a more nested dictionary using the site index as a key too.

Since such a dictionary can be verbose to construct, and since `pydash` is in the `atomate` requirements, an alternative syntax can be:

```
from pydash import set_
response_dict = {}
```

and then, later, when you want to use this you can use:

`_set(response_dict, [key, 'Vdn_site', i], val_to_set)`  # returns {key: 'Vdn_site': i: {val_to_set}

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -775,6 +776,590 @@ def run_task(self, fw_spec):
...
+                        response_dict[rkey]['Ndn_site'+str(i)].append(n_dn)
+                        response_dict[rkey]['Ntot_site'+str(i)].append(n_tot)
+                        response_dict[rkey]['Mz_site'+str(i)].append(m_z)
+
+                        response_dict[rkey]['Vup_site'+str(i)].append(v_up)
+                        response_dict[rkey]['Vdn_site'+str(i)].append(v_dn)
+
+                        response_dict[rkey]['magnetic order'].append(magnet_order)
+
+                except Exception as exc:
```
mkhorton, line 922: This `Exception` is a little too broad; why might the above code fail?

atomate/vasp/firetasks/parse_outputs.py
guymoore13: I'll see what I can do. The reason I had it in that form was because I am setting `rkey` and `is_gs`, in addition to `use_calc`. I'll try to make the logic cleaner that ends at line 880. Thank you!

guymoore13: I changed the logic at lines 861-866 to 

`use_calc = (not relax_nonmagnetic) or (relax_nonmagnetic and int(incar_dict.get("ISPIN"))==2)`

atomate/vasp/firetasks/parse_outputs.py
guymoore13: You're absolutely right, I'll do some tests with that `Exception` removed.

atomate/vasp/firetasks/parse_outputs.py
guymoore13: That's a great point. I think that I was trying to keep them on the same level as `'magnetic order'`. I'll change that.

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -775,6 +776,652 @@ def run_task(self, fw_spec):
         # a builder to put it into materials collection... -computron
         logger.info("Thermal expansion coefficient calculation complete.")
 
+@explicit_serialize
+class HubbardHundLinRespToDb(FiretaskBase):
```
utf, line 780: The code quality looks good here, but it is very difficult to follow along because there is so much code in a single function.

Can you separate out the internal functions and put them in their own file (e..g, `atomate/vasp/analysis/linear_response.py`) and make sure that they have their own docstrings.

Furthermore, can you create new functions for distinct parts of the code. E.g., if you have a lot of code inside a loop, you can move the code inside the loop to a new function and just call the function in the loop.

In general, each function shouldn't be more than say 50 lines long otherwise they get unwieldy.

atomate/vasp/workflows/base/hubbard_hund_linresp.py
utf: Has this been resolved?

atomate/vasp/workflows/base/hubbard_hund_linresp.py
utf: Can you fix this too.

atomate/vasp/workflows/base/hubbard_hund_linresp.py
```diff
@@ -0,0 +1,655 @@
...
+__email__ = "gmoore@lbl.gov"
+__status__ = "Production"
+__date__ = "February 2020"
+
+__hubbard_hund_linresp_wf_version__ = 0.0
+
+module_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)))
+
+
+def get_wf_hubbard_hund_linresp(
```
utf, line 50: There is a similar situation here. This function is too long and should be refactored into separate functions. As it stands it is really difficult to follow along.

utf (commented):
Hi @guymoore13,

In general I think this looks really nice and will be an excellent addition to atomate. To improve the readability of the code, your long functions will need to be refactored into smaller, more focused sub functions. I've written more information in the individual comments.

Best,
Alex

mkhorton:
@Zhuoying this one should also be ready to merge, it went through a few rounds of review with @utf.

Zhuoying:
@guymoore13 Thanks for chunking the codes into small functions. 
It seems the CircleCI failed because of a timeout issue. So I will merge it for now.
@mkhorton @utf Thanks for the review.

guymoore13:
Thank you very much for merging @Zhuoying. And thank you @mkhorton and @utf again for all of your helpful feedback in the process.

PR #415, opened 2020-06-18
Do not store volumetric data by default

mkhorton:
## Summary

We were storing volumetric data by default under the same logic as storing band structure data. These were only stored if explicitly generated. However, given typical database size restrictions, it would be better to turn this off by default and enable on-demand.


utf:
Ok, I support making this change. I'd like to hear if anyone has any objections before merging.

Tagging @computron @bocklund @JaGeo ...

PR #437, opened 2020-07-31
Added extra_outcar_read argument to VaspDrone and VaspToDb.

AndrewBogdan:
## The Issue

Currently VaspDrone serializes OUTCARs with `Outcar.as_dict()`, which does't actually retrieve all the information from the OUTCAR.

In [pymatgen's documentation](https://pymatgen.org/pymatgen.io.vasp.outputs.html?highlight=outcar#pymatgen.io.vasp.outputs.Outcar):

> One can then call a specific reader depending on the type of run being performed. These are currently: read_igpar(), read_lepsilon() and read_lcalcpol(), read_core_state_eign(), read_avg_core_pot().

As far as I can tell, nowhere in atomate are these methods called, except for in the comment (from `atomate.vasp.fireworks.polarization.LcalcpolFW`):

    # Note, Outcar must have read_lcalcpol method for polarization
    # information to be processed. ...assuming VaspDrone will automatically
    # assimilate all properties of the Outcar.

Which, although I don't have the datafiles to test it, I don't believe it does.

## Changes

I added an argument `extra_outcar_read` to VaspDrone and VaspToDb which takes a list of strings. It accepts the following strings:
- `"igpar"`: Call `Outcar.read_igpar()`, storing it under `outcar_dict["igpar"]`.
- `"lepsilon"`: Call `Outcar.read_lepsilon()`, storing it under `outcar_dict["lepsilon"]`.
- `"lcalcpol"`: Call `Outcar.read_lcalcpol()`, storing it under `outcar_dict["lcalcpol"]`.
- `"core_state_eigen"`: Call `Outcar.read_core_state_eigen()`, storing it under `outcar_dict["core_state_eigen"]`.
- `"avg_core_poten"`: Call `Outcar.read_avg_core_poten()`, storing it under `outcar_dict["avg_core_poten"]`

I also edited LcaclpolFW to use `extra_outcar_read=["lcalcpol"]` in its VaspToDb task, although, as I said earlier, I don't have the proper datafiles to test if it works. I _did_ test it with `extra_outcar_read=["core_state_eigen"]`, as that's what prompted me to make this edit. It appears to work fine.

### DIFF
--- a/atomate/vasp/drones.py
+++ b/atomate/vasp/drones.py
@@ -89,7 +89,8 @@ class VaspDrone(AbstractDrone):
                  parse_bader=bader_exe_exists, parse_chgcar=False, parse_aeccar=False,
                  parse_potcar_file=True,
                  store_volumetric_data=STORE_VOLUMETRIC_DATA,
-                 store_additional_json=STORE_ADDITIONAL_JSON):
+                 store_additional_json=STORE_ADDITIONAL_JSON,
+                 extra_outcar_read=None):
         """
         Initialize a Vasp drone to parse vasp outputs
         Args:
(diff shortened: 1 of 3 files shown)


mkhorton:
We could possibly change the pymatgen as_dict() functionality too, I'm not sure which might be preferred. In general, we only get information from the OUTCAR that is absent from the vasprun.xml but I'm not familiar with these specific options to say.

AndrewBogdan:
> We could possibly change the pymatgen as_dict() functionality too, I'm not sure which might be preferred. In general, we only get information from the OUTCAR that is absent from the vasprun.xml but I'm not familiar with these specific options to say.

That sounds like a better idea, as then Outcar.as_dict() would behave as you'd expect.

I looked at the source for Outcar, it appears that they automatically use and serialize `read_lcalcpol` and `read_lepsilon` if possible, but not `read_core_state_eigen`, `read_igpar`, or `read_avg_core_pot`. I'll edit that and make a pull request and see what they say. I was mistaken to say that LcaclpolFW wouldn't work, my bad!

Thank you very much, and sorry for the trouble!

mkhorton:
Hi @AndrewBogdan , I was just following up on this since I realised it had been open a while, did you make the corresponding pymatgen change for this PR?

Zhuoying:
Hi @AndrewBogdan, just a quick follow-up since it has been open for a pretty long period. 
I will close this PR for now. If you are still working on this PR and submit a resolved version, please reply to me and I will reopen this PR for you. Thanks!

PR #454, opened 2020-09-10
Change SCAN workflow to 2-Firework model

rkingsbury:
## Summary

Work in progress updates to the Materials Project SCAN workflow.

* Change SCAN workflow to 2-firework model
* Change the GGA pre-optimization step from PBE to PBESol
* Eliminate passing of the WAVECAR between the GGA and metaGGA steps
* misc other fixes


atomate/vasp/fireworks/core.py
```diff
@@ -131,232 +129,126 @@ def __init__(
...
-            t.append(CopyFiles(from_dir=vdw_kernel_dir))
+        if parents:
+            # Copy the CHGCAR from previous calc (usually PBE)
+            if prev_calc_dir:
+                t.append(CopyVaspOutputs(calc_dir=prev_calc_dir, contcar_to_poscar=True, additional_files=["CHGCAR"]))
+            elif prev_calc_loc:
+                t.append(CopyVaspOutputs(calc_loc=prev_calc_loc, contcar_to_poscar=True, additional_files=["CHGCAR"]))
+            else:
+                raise UserWarning("You specified parent Firework but did not provide its location. Set "
+                                  "prev_calc_dir or prev_calc_loc and try again.")
```
rkingsbury, line 201: @utf can you help me understand whether this is the correct way to pass a CalcLoc between two fireworks? Specifically, I'm confused about `prev_calc_loc`. The docstring for `CopyVaspOutputs` says

> 
> calc_loc (str OR bool): if True will set most recent calc_loc. If str
>             search for the most recent calc_loc with the matching name

Does that mean that if set to True, it will get the most recent calc_loc in the whole database? (i.e., whatever firework happens to have run just before this one) or the most recent calc_loc in the workflow?

I just want to make sure this code is robust and will always link the right fireworks together. Thanks!

rkingsbury:
@utf All dev tests should pass once our `InputSet` changes are merged https://github.com/materialsproject/pymatgen/pull/1952

atomate/vasp/workflows/base/library/SCAN_optimization.yaml
```diff
@@ -1,4 +1,16 @@
...
+    vasptodb_kwargs:
+      task_fields_to_push: {"gga_bandgap": "calcs_reversed.0.output.bandgap"}
+      parse_bader: False
+- fw: atomate.vasp.fireworks.core.ScanOptimizeFW
+  params:
+    name: "SCAN structure optimization"
+    parents: 0
+    prev_calc_loc: True
+    vasptodb_kwargs:
+      parse_bader: False
```
rkingsbury, line 16: @utf @mkhorton @shyamd do you have any feedback on the respective FW names here?

mkhorton: I don't know that "for SCAN" is necessary, I would change the first firework to be "PBEsol structure optimization" (lower-case s)... under the logic that this way it states what it is, rather than what it's for, because there's zero SCAN specific stuff in the first firework.

rkingsbury:
@utf It looks like this is not passing dev tests. The failures are coming from `test_write_vasp` and the FerroElectric workflow, which I don't believe are related. Please let me know if I'm wrong though.

utf:
Hi @rkingsbury, would you be able to pull the latest changes from master into this branch? I'd like to see if the tests are still failing.

Also, is this still WIP?

rkingsbury:
@utf done. As to the WIP status, I wanted to confirm with you that I'm passing the prev_calc_loc correctly (see question above). Also, if you can think of any edge cases that we should add tests for I'd be happy to do that. Otherwise, I think it makes sense to consider this ready to go.

I expect there will be a few iterative improvements to come, but in the interest of keeping atomate in sync with pymatgen, I think it would make sense to merge this so that tests pass and the workflow functions. I do not expect the workflow structure (the 2-firework model, task names, etc.) to change. What may change are a few parameters in the `InputSet` and possibly the way that the `InputSet` determines `KSPACING`.

I've written the tests in such a way that we can make some changes to the `InputSet` parameters without causing failures here. If we have to make changes in pymatgen that cause test failures in atomate, I think we can handle those via a future, coordinated set of PRs. Happy to hear your thoughts on all this.

rkingsbury:
Let me take another look at these latest failures. All tests pass on my local machine so I'll have to investigate what's happening.

rkingsbury:
OK @utf actually the only dev test that's failing now is the `TestFerroelectricWorkflow`.

utf:
Thanks for looking into those failing tests. The TestFerroelectricWorkflow is not failing on master, so it must be something in this PR that is causing it to fail.

Do you have any idea what that could be?

rkingsbury:
Hmm...the failure is coming from `MPStaticSet.from_prev_calc()`, which raises `ValueError: File contains incompatible functionals!` when loading POTCARs. I diff'ed that class in pymatgen master vs. the previous release (v2020.09.14, which is presumably what master is running against) and I don't see any major changes other than some formatting. I can't think of why anything in this PR would affect `MPStaticSet`. Will keep investigating.

rkingsbury:
I have a strong suspicion this is related to the handling of `potcar_spec` within `MPStaticSet.override_from_prev_calc()`. In this PR, I added `potcar_spec` support to `CopyVaspOutputs` but it isn't clear to me why that would have caused this problem. Anyway I've temporarily disabled the potcar_spec powerup for the ferroelectric workflow to see if that changes anything. Does the test system have POTCARs installed?

rkingsbury:
@utf I think I understand this now but I'm still at a loss for why this PR triggered the problem. 

the FerroElectric workflow uses `MPStaticSet`.from_prev_calc()`, which calls `override_from_prev_calc`, which calls `get_vasprun_outcar()`, which calls `VaspRun()` which by default tries to parse the POTCAR files in the calculation directory. The `use_potcar_spec` powerup we have in atomate does not disable this check.

It seems like the appropriate fix will be for the `use_potcar_spec` powerup to pass a `potcar_spec` type kwarg all the way through this call stack in pymatgen. So, we'll need to do another pymatgen PR for that.

What are your thoughts on why this test is still passing in master and how we should fix it?

```
  File "/Users/ryan/miniconda3/envs/mp2/lib/python3.8/site-packages/fireworks/core/rocket.py", line 262, in run
    m_action = t.run_task(my_spec)
  File "/Users/ryan/atomate/atomate/vasp/firetasks/write_inputs.py", line 385, in run_task
    vis = MPStaticSet.from_prev_calc(
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/sets.py", line 1217, in from_prev_calc
    return input_set.override_from_prev_calc(prev_calc_dir=prev_calc_dir)
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/sets.py", line 1178, in override_from_prev_calc
    vasprun, outcar = get_vasprun_outcar(prev_calc_dir)
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/sets.py", line 3012, in get_vasprun_outcar
    Vasprun(vsfile, parse_dos=parse_dos, parse_eigen=parse_eigen),
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/outputs.py", line 352, in __init__
    self.update_potcar_spec(parse_potcar_file)
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/outputs.py", line 931, in update_potcar_spec
    potcar = self.get_potcars(path)
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/outputs.py", line 901, in get_potcars
    potcar = get_potcar_in_path(os.path.split(self.filename)[0])
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/outputs.py", line 885, in get_potcar_in_path
    pc = Potcar.from_file(os.path.join(p, fn))
  File "/Users/ryan/pymatgen/pymatgen/io/vasp/inputs.py", line 2245, in from_file
    raise ValueError("File contains incompatible functionals!")
ValueError: File contains incompatible functionals!```

**Comment by utf** [comment]
Hi @rkingsbury,

I think the changes to `CopyVaspOutputs` and the corresponding update in `use_potcar_spec` are responsible. I'm guessing that in the master branch, the Ferroelectric workflow:
1. Generates a POTCAR.spec file not POTCAR
2. The `CopyVaspOutputs` doesn't copy POTCAR.spec.
3. The Vasprun parsing can't find a POTCAR/POTCAR.spec so skips parsing it (which is perfectly fine)
4. Everything else proceeds as usual.

In this branch, the POTCAR.spec is copied over in step 2 and then cannot be parsed in step 3. I think a better fix would be to edit the `Vasprun.get_potcars` function. Specially, I'd edit make it exclude potcar files with ".spec" in their filename. So edit line 884 in `pymatgen/io/vasp/outputs.py`:

```python
                if fn.startswith('POTCAR'):
```

to be:

```python
                if fn.startswith('POTCAR') and ".spec" not in fn:
```

Would you be able to try making this edit and seeing if it fixes the issues?

I'll start looking over the rest of the PR and respond to your other comments.

rkingsbury:
Thanks @utf! I definitely needed a 2nd pair of eyes on this. Your fix, in conjunction with passing an extra kwarg to `use_potcar_spec`, resolved the test failures. FYI I'm going to force push to erase a couple of commits related to vasptodb_kwargs that were "barking up the wrong tree" so to speak.

rkingsbury:
See https://github.com/materialsproject/pymatgen/pull/1972

rkingsbury:
@utf can you please trigger a re-run of the dev tests? Now that the pymatgen fix is merged I think they should pass. Fingers crossed anyway!

utf:
@rkingsbury The dev tests are now passing, thanks! I'll try and look over the rest of this PR today.

atomate/vasp/fireworks/core.py
utf: Hi @rkingsbury 

The idea is that there is a calc_locs list in the firework spec that can include the calculation location of the previous fireworks. This list is specific to a workflow rather than the whole database. The calc_locs list is appended by each firework that has finished running, and then passed to the next calculation. The latest calculation run will be the last calc loc in the list.

I think the logic here should follow the same process as the StaticFW. In particular, a firework may be initialised without parents but with `prev_calc_dir` set.

I've passed the logic of StaticFW below:

```python
        if prev_calc_dir:
            t.append(CopyVaspOutputs(calc_dir=prev_calc_dir, contcar_to_poscar=True))
            t.append(WriteVaspStaticFromPrev(other_params=vasp_input_set_params))
        elif parents:
            if prev_calc_loc:
                t.append(
                    CopyVaspOutputs(calc_loc=prev_calc_loc, contcar_to_poscar=True)
                )
            t.append(WriteVaspStaticFromPrev(other_params=vasp_input_set_params))
        elif structure:
            vasp_input_set = vasp_input_set or MPStaticSet(
                structure, **vasp_input_set_params
            )
            t.append(
                WriteVaspFromIOSet(structure=structure, vasp_input_set=vasp_input_set)
            )
        else:
            raise ValueError("Must specify structure or previous calculation")
```

To me, there is also a potential issue with the StaticFW logic: You could create a StaticFW with no parents and `prev_calc_loc=True` but then add parents added later (using `Worfklow.append_wf`). In that case, the previous calculation location will not be used.

IMO, it would be better to just throw a warning if `prev_calc_loc` is set but `parents=None`.

rkingsbury: Thanks @utf. I started with that logic but was trying to streamline it a bit, because there quite a few instructions that need to be executed when there is a previous calc, but not otherwise (in the above example, `t.append(WriteVaspStaticFromPrev(other_params=vasp_input_set_params))` is repeated; in my case there is more than that. 

I've updated my code to preserve the logic above, use a flag variable to avoid repeating all those instructions, and to raise a warning if `prev_calc_loc` is set but `parents` is not.

atomate/vasp/firetasks/write_inputs.py
```diff
@@ -289,34 +290,49 @@ def run_task(self, fw_spec):
 
 
 @explicit_serialize
-class UpdateScanRelaxBandgap(FiretaskBase):
+class WriteScanRelaxFromPBE(FiretaskBase):
```
utf, line 293: I wonder if this could be called something else because PBE isn't used. What about `WriteScanRelaxFromGGA`?

atomate/vasp/firetasks/write_inputs.py
```diff
@@ -289,34 +290,49 @@ def run_task(self, fw_spec):
...
 
     """
-    optional_params = ["override_default_vasp_params", "potcar_spec"]
+    optional_params = ["vasp_input_set_params", "potcar_spec"]
 
     def run_task(self, fw_spec):
 
-        kwargs = self.get("override_default_vasp_params")
         potcar_spec = self.get("potcar_spec", False)
+        vasp_input_set_params = self.get("vasp_input_set_params") or {}
```
utf, line 316:
```suggestion
        vasp_input_set_params = self.get("vasp_input_set_params", {})
```

atomate/vasp/firetasks/write_inputs.py
```diff
@@ -289,34 +290,49 @@ def run_task(self, fw_spec):
...
+
+        # update the bandgap based on output from the previous calculation,
+        # unless the user specified a bandgap via vasp_input_set_params
+        if not vasp_input_set_params.get("bandgap"):
```
utf, line 320: If a bandgap of zero is passed, this will evaluate to `False`. I'm not sure if someone would actually do that but maybe change to:
```suggestion
        if vasp_input_set_params.get("bandgap") is None:
```

atomate/vasp/firetasks/write_inputs.py
```diff
@@ -289,34 +290,49 @@ def run_task(self, fw_spec):
...
+            # First look for the gga_bandgap key in the FW spec, to save parsing time
+            if fw_spec.get("gga_bandgap"):
```
utf, line 322: Same here...
```suggestion
            if fw_spec.get("gga_bandgap") is not None:
```

atomate/vasp/fireworks/core.py
```diff
@@ -131,232 +129,126 @@ def __init__(
...
-        # gzip the output
+        # Run VASP
+        t.append(RunVaspCustodian(vasp_cmd=vasp_cmd, auto_npar=">>auto_npar<<", gzip_output=False))
+        t.append(PassCalcLocs(name=name))
+        # Parse
+        t.append(VaspToDb(db_file=db_file, **vasptodb_kwargs))
+        # Delete the VdW kernel
+        t.append(DeleteFiles(files=["vdw_kernel.bindat"]))
+        # zip the output (don't rely on custodian to do it)
         t.append(GzipDir())
```
utf, line 266: Can you get rid of this now and revert to `RunVaspCustodian(... gzip_output=True)`?

utf (commented):
Hi @rkingsbury, this is looking in good shape. I just had a few minor code review comments.

Can you clarify, it looks like your workflow tests don't actually check the workflow completed, rather just that the input files are written correctly. If thats the case, it would be good to have some additional checks at the end of each test to ensure that you can get some data from the task database and to ensure that all fireworks completed without error.

You can see an example in the Ferroelectric workflow tests. I.e., the at the end it has:

```
        wf = self.lp.get_wf_by_fw_id(fw_id)
        is_completed = [s == "COMPLETED" for s in wf.fw_states.values()]
        self.assertTrue(all(is_completed))
```

atomate/vasp/firetasks/parse_outputs.py
```diff
@@ -97,6 +97,7 @@ def run_task(self, fw_spec):
                           parse_dos=self.get("parse_dos", False),
                           parse_potcar_file=self.get("parse_potcar_file", True),
                           bandstructure_mode=self.get("bandstructure_mode", False),
+                          parse_bader=self.get("parse_bader", BADER_EXE_EXISTS),
```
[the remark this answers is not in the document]
rkingsbury: I believe it is turned on by default as long as bader.exe is installed. I've had problems with it either taking a long time or failing, and causing an otherwise good calculation not to be parsed, so I wanted to be able to disable bader parsing in my workflow definition. I'm going to add some lines about it to the docstring which I realize I forgot to do.

atomate/vasp/firetasks/write_inputs.py
rkingsbury: Done. `WriteScanRelaxFromPrev`.

atomate/vasp/fireworks/core.py
rkingsbury: I'd prefer to leave this as is, so that the workflow can be run without Custodian and still behave exactly the same way. To me, it's much easier to understand and work with if custodian ONLY handles error correction while all other tasks (job construction, parsing, zipping) are explicitly defined in atomate. I don't see a reason why removing custodian should cause the output not to be .gzipped, for example.

atomate/vasp/firetasks/parse_outputs.py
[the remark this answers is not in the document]
utf: Ok, it might be good to change this to an env check variable at some point.

atomate/vasp/fireworks/core.py
utf: Ok, that makes sense!

utf:
I think this looks ready to go. We can merge once a new pymatgen version is released (politely paging @mkhorton).

rkingsbury:
Thanks for your attention to it @utf!

atomate/vasp/firetasks/write_inputs.py
```diff
@@ -289,34 +290,49 @@ def run_task(self, fw_spec):
...
 
     """
-    optional_params = ["override_default_vasp_params", "potcar_spec"]
+    optional_params = ["vasp_input_set_params", "potcar_spec"]
 
     def run_task(self, fw_spec):
 
-        kwargs = self.get("override_default_vasp_params")
         potcar_spec = self.get("potcar_spec", False)
+        vasp_input_set_params = self.get("vasp_input_set_params", {})
```
utf, line 316: I think I asked you to change this previously but I've realised that it will cause an error if vasp_input_set_params is set with a default value of `None`.

Can you change it back to:
```suggestion
        vasp_input_set_params = self.get("vasp_input_set_params") or {}
```

utf:
Hi @rkingsbury, can you update the project requirements to the latest pymatgen version. Then hopefully we can merge.

PR #476, opened 2020-10-23
Change source of bandgap in SCAN workflow logic

rkingsbury:
It has come to our attention that `Vasprun.get_band_structure().get_band_gap()["energy"]` can return an incorrect band gap in some cases. The bandgap is used to determine k-point density and smearing settings for SCAN calculations in `WriteScanRelaxFromPrev`. 

To make this firetask more robust, this PR switches the source of the bandgap information to `Vasprun.eigenvalue_band_properties[0]`.

### DIFF
--- a/atomate/vasp/firetasks/write_inputs.py
+++ b/atomate/vasp/firetasks/write_inputs.py
@@ -325,7 +325,7 @@ class WriteScanRelaxFromPrev(FiretaskBase):
             else:
                 parse_potcar_file = not potcar_spec
                 vasprun = Vasprun("vasprun.xml", parse_potcar_file=parse_potcar_file)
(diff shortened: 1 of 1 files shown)


rkingsbury:
@mkhorton FYI

utf:
Thanks!

Does it make sense to fix `BandStructure.get_band_gap()` to have the same logic as `Vasprun.eigenvalue_band_properties`?

rkingsbury:
I'll defer to @mkhorton on that one.

mkhorton:
Yes, but it would require storing occupancies in the band structure object I believe, i.e. not a small change.

mkhorton:
At the least, we probably should store both in the task doc though, so we can query for when they disagree substantially.

PR #479, opened 2020-10-27
Small maggma store changes

jmmshn:
## Small maggma store changes

Creature comfort changes:

* Allow "maggma_store_prefix" to be defined in the db.json file to avoid contamination of dbs
* Made sure the documents are searchable by "task_id"


utf:
Thanks!

PR #482, opened 2020-10-29
Mongo URI parsing for CalcDb

jmmshn:
## Added treatment for URIs for creating the task database

This was somewhat needed for the sharded databases for large emmet-cli ingestion runs. 
But should also make the configuration cleaner in general. 

You can now set a `host_uri`  in the db_file if a database name is present it will overwrite the database given to the init.
Instead of parsing the uri all unset variables are kept as None.


atomate/utils/database.py
```diff
@@ -41,6 +42,7 @@ def __init__(self, host, port, database, collection, user, password, maggma_stor
             collection: the collection were the parsed dictionaries will be stored
             user: MongoDB authentication username
             password: MongoDB authentication password
+            host_uri: If the uri disgination of the mongodb is provided, other authentication information will be ignored
```
utf, line 45: Typo on designation, also can you sort out this line length too.

atomate/utils/database.py
```diff
@@ -212,15 +240,20 @@ def _get_s3_store(self, store_name):
...
-            host=self.host,
-            port=self.port,
-            username=self.user,
-            password=self.password,
-            key = 'fs_id'
-        )
+        if self.host_uri is not None:
+            index_store_ = MongoURIStore(uri=self.host_uri, database=self.db_name,
+                                         collection_name = f"{self.maggma_store_prefix}_{store_name}_index",
+                                         )
```
utf, line 246: Can you tidy this too pls.
```suggestion
            index_store_ = MongoURIStore(
                uri=self.host_uri,
                database=self.db_name,
                collection_name = f"{self.maggma_store_prefix}_{store_name}_index",
            )
```

utf (commented):
Looks good, can you just tidy the line length and formatting in the places I highlighted.

utf:
Thanks @jmmshn

PR #488, opened 2020-10-29
forgot to set key to fs_id for URI store

jmmshn:
also used `black` on the test file


utf:
You may as well run black on CalcDb and VaspCalcDb too please.

jmmshn:
done

PR #512, opened 2020-11-20
Ion insertion workflow

jmmshn:
## Workflow to automatically insert ions to a host material

* Added `power_up_by_kwargs` to allow powerups for dynamically created workflows
* Added tasks to insert ions


atomate/vasp/fireworks/core.py
```diff
@@ -325,7 +345,18 @@ def __init__(
...
+                structure, **vasp_input_set_params
+            )
+            t.append(
+                WriteVaspFromIOSet(
+                    structure=structure,
+                    vasp_input_set=vasp_input_set,
+                    spec_structure_key="prev_calc_structure",
+                )
+            )
+        elif prev_calc_dir:
```
jmmshn, line 359: This allows the spec to be used to pass the structure.  All the other changes in this file are formatting.

jmmshn:
OK so this test works on my computer and I have no idea why it's not working here.

jmmshn:
OK nvm, I just needed to add scikit-image to the requirement.
I think this is OK for review now.  Thanks!

atomate/vasp/powerups.py
```diff
@@ -939,3 +940,35 @@ def use_fake_lobster(original_wf, ref_dirs, params_to_check=None):
...
+
+    As an example:
+        power_up_by_kwargs( "add_additional_fields_to_taskdocs" : {
+                                                                "update_dict" : {"foo" : "bar"}
+                                                                }
+        )
+    """
+    for k, v in kwargs.items():
+        wf = local_names[k](wf, **v)
+    return wf
```
jmmshn, line 974: @utf, @mkhorton: (roping matt in b/c I also discussed this with him)
This is kludgey part for passing the powerups to dynamically created fws.
The `local` needs to be called above so that it only loads and import time.

atomate/vasp/firetasks/electrode_tasks.py
```diff
@@ -0,0 +1,311 @@
...
+            for fw in wf.fws:
+                fw.spec.update({k: fw_spec.get(k, None)})
+
+
+@explicit_serialize
+class AnalyzeChgcar(FiretaskBase):
+    """
+    Process the charge density from a particular task.
+    fw_spec requires:
+    - "base_task_id" : the task_id for the charge density that is being analyzed
```
utf, line 60: Is there any reason you are passing these variables via the FW spec, rather than using `required_params` and `optional_params`?

If not, could you convert the Firetask to use those instead?

atomate/vasp/firetasks/electrode_tasks.py
```diff
@@ -0,0 +1,311 @@
...
+            }
+        )
+
+
+@explicit_serialize
+class GetInsertionCalcs(FiretaskBase):
+    """
+    Using a list of insertion sites and a base structure, get a new set of calculations
+
+    fw_spec requires:
```
utf, line 125: Similarly for this Firework? Why not required and optional params?

atomate/vasp/firetasks/electrode_tasks.py
```diff
@@ -0,0 +1,311 @@
...
+            fpos = isite
+            inserted_structure.insert(0, working_ion, fpos, properties={"magmom": 1.0})
+            additional_fields = {"insertion_fpos": fpos, "base_task_id": base_task_id}
+
+            # Create new fw
+            fw = OptimizeFW(
+                inserted_structure,
+                name=f"structure optimization-{itr}",
+                override_default_vasp_params={
+                    "user_incar_settings": {"NSW": 9999}
```
utf, line 159: 😨 This is an enormous number. Do the insertions really need 10,000 steps? I'm worried that if VASP can't optimise the structure, it will burn computer time with the atoms bouncing around without converging.

atomate/vasp/firetasks/electrode_tasks.py
```diff
@@ -0,0 +1,311 @@
...
+                },  # structure are rough guesses
+                **optimizefw_kwargs,
+            )
+            fw.tasks[-1]["additional_fields"].update(
+                additional_fields
+            )  # easy way to update a field
+
+            pass_dict = {
+                "structure": ">>output.ionic_steps.-1.structure",
+                "energy": "a>>final_energy",
```
utf, line 169: should the "a" be there?

atomate/vasp/firetasks/electrode_tasks.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@explicit_serialize
+class CollectInsertedCalcs(FiretaskBase):
+    """
+    Aggregate the outputs of the inserted calculations, take the lowest energy completed calculation
+    and pass it on for charge density analysis
+
+    Required keys in fw_spec:
+    - "base_structure": the base structure that the ion will be inserted on
```
utf, line 207: Same with this firetask?

atomate/vasp/firetasks/electrode_tasks.py
```diff
@@ -0,0 +1,311 @@
...
+        # Get the new structures
+        return FWAction(update_spec={"optimal_structure": best_res["structure"]})
+
+
+@explicit_serialize
+class SubmitMostStable(FiretaskBase):
+    """
+    For the best structure submit the Static WF
+
+    Required keys in fw_spec:
```
utf, line 262: Also with this Firetask?

atomate/vasp/powerups.py
utf: I think this looks ok to me. It isn't the cleanest but will do!

atomate/vasp/test_files/insertion_wf/loop.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
utf, line 1: Does this still need to live here?

utf (commented):
Hi @jmmshn. This PR is looking good. My only concern is that you've used the fw_spec to pass all optional and required parameters in your firetasks. I'm not clear why you've had to do that rather than using the `required_params` and `optional_params` keys that firetasks usually use.

atomate/vasp/firetasks/electrode_tasks.py
jmmshn: So I looked at this one specifically, and even here since this relies on `task_fields_to_push` to work the `base_task_id` have to be stored in fw.spec.

atomate/vasp/firetasks/electrode_tasks.py
jmmshn: oop I think that was introduced during testing. I have set it to a more reasonable value.

atomate/vasp/test_files/insertion_wf/loop.sh
jmmshn: No I will delete this

jmmshn:
Hi @utf I looked through the code and even simple stuff like the task_id have to be parsed via the spec.
So I think it makes sense to just keep the current setup.  I have fixed the other problems and this should be ready to merge after tests pass.

utf:
@jmmshn please can you resolve the conflicts then I can merge.

PR #560, opened 2021-01-07
Improved handling of QChem scratch directories, saving scratch files, and constructing opt trajectories

samblau:
## Summary

This PR goes hand-in-hand with recently merged changes in Custodian which substantially improve the management of QChem's environment variables responsible for controlling the location and saving of scratch files. It also makes the linked frequency flattening optimizer the default and improves the drone's construction of an optimization trajectory for the task doc.


samblau:
Will work on fixing tests tomorrow

utf:
Thanks @samblau! I'm not familiar with the qchem modules but the code looks good from a pass over.

PR #598, opened 2021-03-01
Q-Chem electron density critical point analysis via Critic2

samblau:
## Summary

Adds the ability to analyze Q-Chem molecular electron density critical points with the Critic2 package (where Q-Chem prints the electron density cube file that Critic2 analyzes). This has been incorporated into our production molecular simulation procedure in order to improve the identification of chemical bonds in complex open shell structures, particularly those with coordinating metal ions. Note that Critic2 must be compiled/installed in order to be run, so some emulation was required for testing purposes. Also note that most of the 40 files changed are just associated with tests - there are only nine files with new or updated code. 

* New RunCritic2 firetask and test
* Critic2 processed output parsing added to the QChem drone along with a new associated test
* CubeAndCritic2 firework added to core along with a new associated test
* Critic2 emulation capabilities added to Powerups
* New FFopt_and_Critic workflow and test


samblau:
@utf  - this is ready for review / merging. Thanks!

utf:
Hi @samblau, thanks very much for this! I'll hopefully have a chance to review this afternoon.

samblau:
@utf, much appreciated!

mkhorton:
I've had a read over the Critic2 code here, it seems fine aside from small PEP8 issue (`C2` variable).

I would recommend that this is better incorporated in pymatgen both because it's actively tested in pymatgen (afaik, Critic2 is not installed with atomate) and to keep the atomate implementation leaner for future atomate changes.

If incorporated into pymatgen, I suggest the cleanest way to do this would be to edit `Critic2Caller` class, and make the main `__init__` and much simpler `def __init__(self, input_script)`, rename the current `__init__` to be a classmethod `def call_from_chgcar(...)` and add a new classmethod `def call_from_cube(...)`.

I think this would be most beneficial to others who might wish to run Critic2 on cube files too, rather than burying the functionality inside a FireTask.

Since I'm not an atomate maintainer, I'll leave the final call on this with @utf.

samblau:
Your points are very reasonable. However, if possible, I would greatly appreciate if this could be merged in as-is (or maybe once I fix the PEP8 issue?) and then an issue could be opened so that I could work towards a better overall implementation of the Critic2 infrastructure in the future. We're about to submit a paper to Sci Dat, and the code in this PR is what we've used for all production Critic2 analysis. It should have not taken me this long to PR our Critic2 code into Master, and I recognize that it needs to be reworked and improved, but we'd really like it to be in Master for the paper submission. Then I can work on improving stuff in the coming months. Thoughts? Thanks!

utf:
I agree with @mkhorton that having code to run Critic2 in both atomate and pymatgen should be avoided. It seems like it should be fairly simple to make the changes Matt is suggesting to the `Critic2Caller` class in pymatgen. I'm conscious that if I merge this now, making the changes to pymatgen will be forgotten and we will be stuck with the duplicated code forever.

@samblau, how about I merge this into a new branch on this repository, that way people can still use the functionality by switching to that branch without having to clone your version of atomate. Once the changes are made in pymatgen I can merge into master. We did this recently for one of Jimmy's Pull requests and that has worked very well.

atomate/qchem/firetasks/critic2.py
```diff
@@ -0,0 +1,178 @@
...
+            tmp = atom["cartesian_coordinates"]
+            coords = []
+            for jj,val in enumerate(tmp):
+                coords.append((val+centering_vector[jj])*bohr_to_ang)
+            if str(molecule[ii].specie) != specie:
+                raise RuntimeError("Atom ordering different!")
+            if molecule[ii].distance_from_point(coords) > 1*10**-5:
+                raise RuntimeError("Atom position "+str(ii)+" inconsistent!")
+
+        assert cp_loaded["critical_points"]["number_of_nonequivalent_cps"] == cp_loaded["critical_points"]["number_of_cell_cps"]
```
utf, line 132: Can you change this to:

```python
if (
    cp_loaded["critical_points"]["number_of_nonequivalent_cps"] != 
    cp_loaded["critical_points"]["number_of_cell_cps"]
):
    raise ValueError("error message")
```

As assertions will be ignored if python is run with optimisations.

atomate/qchem/powerups.py
```diff
@@ -2,15 +2,22 @@
...
+__email__ = "samblau1@gmail.com"
+__status__ = "Alpha"
+__date__ = "11/20/19"
 
 
 def use_fake_qchem(original_wf, ref_dirs, input_file="mol.qin"):
     """
         Replaces all RunQChem commands (i.e. RunQChemDirect, RunQChemCustodian) with RunQChemFake.
         This allows for testing without actually running QChem
+        Also turns on testing=True for RunCritic2 to similarly facilitate testing
```
utf, line 20: Instead implement a RunFakeCritic2 or even just delete the `RunCritic2` firetask.

utf (commented):
Hi @samblau, I've done a more thorough review, see the attached comments.

In general this is a great pull request. Thank you for adding all the tests! My only issue is the `testing` parameter for `RunCritic2`. Instead it would be much cleaner to split this into two Firetasks as I mention in the comments. This won't require any major rewrites of the code so  shouldn't take too much effort.

samblau:
Hi @utf, thanks for the feedback! Those all seem like very reasonable suggestions - I'll work on them shortly. I knew the "testing" input was bad form, but I hadn't thought to split it into two firetasks. Good call.

However, I would prefer to avoid having this merged into a non-master branch. I would rather hold this PR open while taking care of the necessary pymatgen changes. But after spending some time looking at the `Critic2Caller`, I'm worried that this might be more difficult than expected because so many assumptions are being made that the analysis is being performed on a structure instead of a molecule. I will try and find some time to chat with @mkhorton about this soon.

mkhorton:
>  But after spending some time looking at the Critic2Caller, I'm worried that this might be more difficult than expected because so many assumptions are being made that the analysis is being performed on a structure instead of a molecule.

To clarify the suggestion, `Critic2Caller.__init__` can be _very_ minimal, e.g. something like this:

```
def __init__(input_script):

    # code below copied from current init
    with ScratchDir(".") as temp_dir:

            os.chdir(temp_dir)

            with open("input_script.cri", "w") as f:
                f.write(input_script)

            args = ["critic2", "input_script.cri"]
            rs = subprocess.Popen(args, stdout=subprocess.PIPE, stdin=subprocess.PIPE, close_fds=True)

            stdout, stderr = rs.communicate()
            stdout = stdout.decode()

            if stderr:
                stderr = stderr.decode()
                warnings.warn(stderr)

            if rs.returncode != 0:
                raise RuntimeError("critic2 exited with return code {}: {}".format(rs.returncode, stdout))

            self._stdout = stdout
            self._stderr = stderr

            if os.path.exists("cpreport.json"):
                cpreport = loadfn("cpreport.json")
            else:
                cpreport = None
            self._cpreport = cpreport

            if os.path.exists("yt.json"):
                yt = loadfn("yt.json")
            else:
                yt = None
            self._yt = yt
```

then we can copy and paste the Chgcar-specific code from the current __init__ into a new class method, `from_chgcar()`, and add a new `from_cube()`.

The code from the FireTask can be put into `from_cube()`. In fact, this code looks so similar I suspect it was copied from the pymatgen code anyway?

The job of `Critic2Caller` is just to get you the JSON you need for your analysis. What you do after that is up to you. The `Critic2Analysis` class does a lot of parsing and checking of the JSON output, this could be adapted to handle molecules too, but that might be OTT for your needs.

samblau:
Setting `self.output = Critic2Analysis()` was one of my big concerns, because Critic2Analysis assumes a structure. But if that's only in `call_from_chgcar`, then we could be in business. I'll be in touch again when I have time to come back and focus on this.

atomate/qchem/powerups.py
samblau: FYI, I went with the second option and now just delete the `RunCritic2` as part of `use_fake_qchem`. It feels a little weird to do something critic-specific in what appears to be a qchem-specific function, but since Critic2 is run in the same directory as the QChem calculation, and thus the Critic2 outputs are "generated" via emulation along with the QChem outputs, this is definitely the simplest solution.

atomate/qchem/firetasks/critic2.py
samblau: Changed in latest commit

atomate/qchem/firetasks/tests/test_critic2.py
```diff
@@ -0,0 +1,57 @@
...
+        os.remove("bonding.json")
+        shutil.move("bonding_correct.json","bonding.json")
+
+    def test_RunCritic2(self):
+        os.chdir(os.path.join(module_dir, "..", "..", "test_files",
+                                "critic_test_files", "critic_example"))
+        firetask = RunCritic2(
+            molecule=self.mol,
+            cube_file=self.cube_file,
+            testing=True)
```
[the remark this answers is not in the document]
samblau: Changed in latest commit

atomate/qchem/firetasks/critic2.py
```diff
@@ -0,0 +1,178 @@
...
+__maintainer__ = "Samuel Blau"
+__email__ = "samblau1@gmail.com"
+__status__ = "Alpha"
+__date__ = "11/20/19"
+
+
+logger = get_logger(__name__)
+
+@explicit_serialize
+class RunCritic2(FiretaskBase):
```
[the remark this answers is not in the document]
samblau: Changed in latest commit

atomate/qchem/firetasks/critic2.py
```diff
@@ -0,0 +1,178 @@
...
+    Optional params:
+        testing (Bool): Set to true when running tests to avoid actually calling the Critic executable    
+    """
+
+    required_params = ["molecule", "cube_file"]
+
+    optional_params = ["testing"]
+
+    def run_task(self, fw_spec):
+        if fw_spec.get("prev_calc_molecule"):
```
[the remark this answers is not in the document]
samblau: Changed in latest commit

mkhorton:
`units` were added at basically the same time the JSON output was added, I believe in the same thread, so only a few commits apart if that. I asked because `critic2` can change its output units based on an input directive, so I wanted to be sure what we had when we parsed an output file.

samblau:
I think that was juuuuust after I compiled critic before, hah! I recompiled (after first updating gfortran/gcc, which required me first updating brew... sigh) and now it's working fine. Also just fixed the class method problems I was having. Should be submitting a pymatgen PR shortly.

mkhorton:
Fantastic, thanks!

samblau:
@mkhorton @utf All fixed - this is ready for a re-review. Thanks for pushing me to do it the right way :)

utf:
Thanks @samblau, this looks great.

PR #608, opened 2021-03-09
Make pymatgen import statements more specific

stevetorr:
## Summary

The latest update to Pymatgen removed the `__init__.py` file such that import statements drawing from the package e.g. `from pymatgen import Structure` no longer work. There are a few of these across the repo. This PR fixes all of them I could find. Helpfully, this also makes it backwards compatible, since the specificity does not change any functionality.

* All statements of the form `from pymatgen import x` are now made more specific, e.g. `from pymatgen import Structure` is now `from pymatgen.core import Structure`.

### DIFF
--- a/atomate/feff/firetasks/tests/test_tasks.py
+++ b/atomate/feff/firetasks/tests/test_tasks.py
@@ -6,7 +6,7 @@ import shutil
 import unittest
 from glob import glob
 
-from pymatgen import Structure
+from pymatgen.core import Structure
 from pymatgen.io.feff.sets import MPEXAFSSet
 from pymatgen.io.feff.inputs import Paths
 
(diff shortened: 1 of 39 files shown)


utf:
Hi @stevetorr.

Thanks for submitting this PR. It is very much appreciated.

Before we can get this merged, could you also address the following comments:
1. Update the pymatgen version in requirements.txt to the latest version that includes the namespace reorganisation (i.e., pymatgen==2022.0.3). This might expose more issues where the imports are not fully working, so there may be more tests to fix.
2. Also fix any imports in the atomate tests. Particularly, it seems like the the py3devtests are failing due to the error:

```python
atomate/utils/testing.py:13: in <module>
    from pymatgen.settings import SETTINGS
E   ModuleNotFoundError: No module named 'pymatgen.settings'
```

stevetorr:
Thanks , @utf ! Hopefully this is good to go now :)

utf:
Fantastic. Thank you very much.

Issue #576, opened 2021-01-25
Tags are not passed to individual fireworks in QChem workflow


rkingsbury:
This is a bug report to be addressed by internal development. Apologies if it should be redirected to Discourse.

When I create a workflow from `SinglePointFW` or `FrequencyFlatteningOptimizeFW` and assign tags to the workflow, e.g.

```
fw = FrequencyFlatteningOptimizeFW(s)
wf = Workflow([fw], name=wfname, metadata={"tags":["foo","bar"]})
```

The tags are correctly assigned to the `metadata` key of the workflow, but are not passed to the underlying FireWorks. Normally (At least in the VASP workflows), the tags are passed through to the FireWork `spec`. Since queries in `my_fworker.yaml` operate on individual FireWorks, the lack of tagging makes it difficult to find and organize calculations.

If someone can provide guidance on where / how this should be fixed I'd be happy to open a PR.

rkingsbury:
@samblau @espottesmith have you noticed this?

utf:
If you're using VASP, the typical way of adding tags is via the `add_tags` powerup:

```
def add_tags(original_wf, tags_list):
    """
    Adds tags to all Fireworks in the Workflow, WF metadata, as well as
    additional_fields for the VaspDrone to track them later (e.g. all fireworks
    and vasp tasks related to a research project)

    Args:
        original_wf (Workflow)
        tags_list: list of tags parameters (list of strings)

    Returns:
       Workflow
    """

    # WF metadata
    if "tags" in original_wf.metadata:
        original_wf.metadata["tags"].extend(tags_list)
    else:
        original_wf.metadata["tags"] = tags_list

    # FW metadata
    for idx_fw in range(len(original_wf.fws)):
        if "tags" in original_wf.fws[idx_fw].spec:
            original_wf.fws[idx_fw].spec["tags"].extend(tags_list)
        else:
            original_wf.fws[idx_fw].spec["tags"] = tags_list

    # DB insertion tasks
    for constraint in ["VaspToDb", "BoltztrapToDb"]:
        idxs = get_fws_and_tasks(original_wf, task_name_constraint=constraint)
        for idx_fw, idx_t in idxs:
            if "tags" in original_wf.fws[idx_fw].tasks[idx_t]["additional_fields"]:
                original_wf.fws[idx_fw].tasks[idx_t]["additional_fields"][
                    "tags"
                ].extend(tags_list)
            else:
[...]

PR #615, opened 2021-03-18
Move common powerups out of the VASP subdirectory

rkingsbury:
## Summary

Several powerups currently in `atomate.vasp.powerups` are useful and work properly on any type of workflow, not just VASP workflows. Their location inside the VASP directory obscures this functionality for users of other codes (See #576)

This PR moves the following powerups to `atomate.common.powerups` to better reflect their universal nature. In addition, the `add_tags` powerup is modified slightly so that it works consistently for any workflow.
* `add_tags`
* `add_priority`
* `add_namefile`
* `add_additional_fields_to_taskdocs`
* `preserve_fworker`
* `set_execution_options`

Appropriate deprecation warnings are added, and the current functionality of powerups in `atomate.vasp.powerups` is maintained.

There is also a small change to `QChemToDb` that removes unnecessary code, now that add_tags has been added.


atomate/qchem/firetasks/parse_outputs.py
```diff
@@ -77,9 +77,6 @@ def run_task(self, fw_spec):
             output_file=output_file,
             multirun=multirun)
 
-        if "tags" in fw_spec:
-            task_doc.update({"tags": fw_spec["tags"]})
-
```
rkingsbury, line 82: Hey @samblau , in this PR I'm modifying the `add_tags` powerup to be more general and work properly with `QChemToDb` (see #576 . That means the lines removed here will be redundant, and if left in place could result in duplicate tags in the task doc.

I wanted to flag you because IF members of your team are in the habit of directly setting tags in the Firework (i.e. `<Firework>["_spec"]["tags'] = "some_tag"` and then expecting tags to go into the task doc, that functionality will break with this change. Instead, they'd need to use `atomate.common.powerups.add_tags`. Are you OK with that?

As an alternative, I could add some logic that checks to see if 'tags' is listed in 'additional_fields', and if not, passes the `fw_spec` tags through, preserving the current functionality. Please let me know your preference.

utf:
Hi @rkingsbury, thanks for this! Could you move the tests for the powerups you moved to `atomate/common/tests/test_powerups.py` and also run black on the files you edited.

Other than that, this looks great.

atomate/qchem/firetasks/parse_outputs.py
samblau: This is fine with me. FYI @espottesmith

espottesmith: Not an issue for any of my workflows either.

rkingsbury:
Should be all set @utf. I copied the relevant tests to `atomte.common.tests.test_powerups.py` but also left the originals in `atomate.vasp.tests.test_vasp_powerups.py` since the legacy powerups still reside in that file. It's just an extra layer of verification that the deprecated ones continue to work properly until removed.

utf:
Ok, great. Thank you.

PR #618, opened 2021-03-26
ApproxNEB Workflow for Dilute Lattice Limit

acrutt:
This workflow implements the ApproxNEB method published in 2016 (https://doi.org/10.1063/1.4960790) for evaluating the energetics along the migration path of a working ion moving through a host lattice. This method is intended as a more robust and less computationally expensive alternative to Nudged Elastic Band calculations. This version of the workflow is only appropriate for the dilute lattice limit where a single working ion is in a large host supercell structure where little volume change is expected upon inserting the working ion.

There are 4 types of fireworks in this workflow:

1. **HostFW:** relaxes the empty host structure provided and stores the CHGCAR for later use in the EvaluatePathFW
2. **EndPointFW:** inserts a working ion into the relaxed host structure from the HostFW and performs another relaxation. The cell shape and volume is fixed in this relaxation in order to keep the lattice consistent.
3. **EvaluatePathFW:** takes two end point structures and interpolates images that connect the two working ion positions from the end points. The CHGCAR of the empty host structure is used to find the connecting path with the lowest charge density. Dynamically generates fireworks for the images generated.
4. **ImageFW:** Performs a relaxation of the image structures with selective dynamics applied so two atoms are fixed: 1) the migrating working ion and 2) the atom furthest away from the migrating working ion

Here is an example of what a workflow would look like for a MnO2 host material with 4 different Li sites/end point structures. Two different paths are evaluated with 7 images generated for each.
image

Because there is a significant amount of structure manipulation between fireworks, this workflow uses another "approx_neb" MongoDB collection in addition to the typical fireworks collections to track calculation inputs & outputs. Given the complexity of this workflow, input structures are pulled from the approx_neb collection rather than being passed from one parent firework to its child.

FYI @jmmshn and @mkhorton


utf:
This looks fantastic. Thanks you!

mkhorton:
Congratulations on getting this merged @acrutt !

acrutt:
Thank you @mkhorton :)

PR #623, opened 2021-04-01
better handling of dynamic powerups

jmmshn:
Minor rework of how powerups are discovered during dynamic workflows.
- the `powerup_by_kwargs` function now has a sequence of the modules where it will look for powers (similar to a path variable)
- the user can also provide new modules that contain additional powerups using the `add_powerup_module` flag
- If a powerup is found, in any module, apply that powerup and stop looking

### DIFF
--- a/atomate/common/powerups.py
+++ b/atomate/common/powerups.py
@@ -7,7 +7,8 @@ This module defines general powerups that can be used for all workflows
 from atomate.utils.utils import get_fws_and_tasks
 from fireworks import Workflow, FileWriteTask
 from fireworks.utilities.fw_utilities import get_slug
-from inspect import getmembers, isfunction, stack, getmodule
+from inspect import getmembers, isfunction
+import sys
 
 __author__ = "Janine George, Guido Petretto, Ryan Kingsbury"
 __email__ = (
@@ -221,7 +222,7 @@ def set_queue_adapter(
     return original_wf
 
 
-def get_powerup_dict():
+def get_powerup_dict(module):
     """
     Return a complete list of functions in any module where where this `get_powerup_dict` is called.
     Note: this is a quick and dirty way to serialize powerup functions for dynamic workflows.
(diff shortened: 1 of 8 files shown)


jmmshn:
@utf  I don't know why it's github no longer running tests on this PR.
I only see "This branch has no conflicts with the base branch"

utf:
I think there is just a backlog of circleCI jobs waiting to be run. I'll try cancelling the ones that are still running.

jmmshn:
OK I see, I just assumed they automatically canceled once a PR is created

atomate/common/powerups.py
```diff
@@ -219,3 +220,54 @@ def set_queue_adapter(
         original_wf.fws[idx_fw].spec["_queueadapter"] = q
 
     return original_wf
+
+
+def powerup_by_kwargs(
+    original_wf: Workflow,
+    add_powerup_module: List[str] = None,
```
utf, line 227: I don't like this way of specifying additional modules. Can you change this so that:
1. If the powerup name doesn't contain a period then try to import it from the default list of modules ("atomate.vasp.powerups", "atomate.qchem.powerups", "atomate.common.powerups"). If that fails then throw an error saying `"Could not find powerup {}.".format(powerup_name)`
2. If the powerup name contains a period then use everything before the last period as the module path, and everything after the last the period as the powerup name.
3. Make it clear that you can specify a powerup using the full module path in the documentation and that this is the recommended way to use powerups that are not defined in atomate. E.g., `"mypackage.mymodule.powerups.my_powerup"` will use the `my_powerup` powerup imported from `mypackage.mymodule.powerups`.

jmmshn: So the way of using the keys as the names is problematic since Mongo does not allow the keys to have dots.
I have slightly reformatted the way the function name and args are passed and added the error message as you have suggested.

utf (commented):
Looks great. Thank you.

PR #629, opened 2021-04-11
Small patch to add option to disable/enable Bader analysis on VASP ingestion

mkhorton:
## Summary

Bader parsing already exists inside the VASP drone. Currently, it is enabled if `bader` is in your $PATH.

However, this behavior is problematic since some HPC systems automatically do this when you load VASP (e.g. Cori KNL `vasp/5.4.4-knl` module), which may result in a Bader analysis being performed when this was not desired, leading to longer jobs or an additional failure mode.

This PR adds a new config option to explicitly specify whether Bader analysis should be performed.


mkhorton:
Thanks @utf !

utf:
Thanks!

PR #654, opened 2021-05-08
fix pymatgen-diffusion imports

rkingsbury:
It seems that `pymatgen.analysis.diffusion` was removed from pymatgen and into `pymatgen-diffusion` without a deprecation message. Since code from this package is imported in `vasp_powerups.py`, this import problem caused basically all VASP workflows to break.

### DIFF
--- a/atomate/vasp/firetasks/neb_tasks.py
+++ b/atomate/vasp/firetasks/neb_tasks.py
@@ -4,7 +4,7 @@ import shutil
 
 from pymatgen.core import Structure
 from pymatgen.io.vasp import Incar, Kpoints, Poscar, Potcar
-from pymatgen.analysis.diffusion.neb.io import (
+from pymatgen_diffusion.neb.io import (
     MVLCINEBSet,
     get_endpoint_dist,
     get_endpoints_from_index,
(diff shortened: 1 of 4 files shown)


utf:
I think upgrading `pymatgen_diffusion` to `pymatgen-diffusion` as well as upgrading the `pymatgen` version should do the trick. From: https://github.com/materialsvirtuallab/pymatgen-diffusion

> ## Major Update (v2021.3.5)
>
> pymatgen-diffusion is now released as a namespace package pymatgen-analysis-diffusion on PyPI. It should be imported via pymatgen.analysis.diffusion instead pymatgen_diffusion.

utf:
`pymatgen-analysis-diffusion` was already added to requirements.txt but not `setup.py`. I just fixed that.

The remaining issue might just be not using the latest pymatgen version?

rkingsbury:
Hmmm ok, thanks for clarifying. I already have `pymatgen-diffusion` installed and am using the latest pymatgen and atomate from github, but I guess my environment missed replacing `pymatgen-diffusion` with `pymatgen-analysis-diffusion`. Sounds like that will sort it out.

PR #660, opened 2021-05-13
Add common powerup add_metadata

### LINKED ISSUE #659: How to add metadata to workflow?
What's the recommended way of adding top-level metadata to fireworks? With top level, I mean directly accessible as a key-value pair in the entries returned by the tasks collection after calculations have finished.

My use case is adding the MP `material_id` for further analysis and comparison. Doing

```py
wf.metadata["mp_id"] = material_id
```

does not convey the id into the database. Using

```py
wf = add_tags(wf, [material_id])
```

_does_ insert the tags into MongoDB but is suboptimal because all tags end up in a column of lists when converted to a dataframe. A list column is much harder to deal with than a column of primitive types, and it can't be parsed into separate columns with `pd.json_normalize` the way a dict could.

janosh:
Closes #659.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 # Contributing to atomate
+
 We love your input! We want to make contributing to atomate as easy and transparent as possible, whether it's:
+
 * Reporting a bug
 * Discussing the current state of the code
 * Submitting a fix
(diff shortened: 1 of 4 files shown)


janosh:
@utf Ping

utf:
Thanks very much for this PR.

janosh:
Any chance a new version of atomate could be released to make the `add_metadata` powerup available?

computron:
Sorry I missed this! I just released v1.0.1 which should have this in

PR #669, opened 2021-05-25
Insertion Worflow optimizefw_kwargs

acrutt:
Debug GetInsertionCalcs firetasks optimizefw_kwargs so user can provide override_default_vasp_params without conflicting with setting NSW=299 (if NSW is not set by the user)

See Insertion Worflow optimizefw_kwargs Issue for bug description:
[https://matsci.org/t/insertion-worflow-optimizefw-kwargs-issue/36593](https://matsci.org/t/insertion-worflow-optimizefw-kwargs-issue/36593)

### DIFF
--- a/atomate/vasp/firetasks/electrode_tasks.py
+++ b/atomate/vasp/firetasks/electrode_tasks.py
@@ -1,4 +1,5 @@
 import math
+from pydash import get
 
 from fireworks import FiretaskBase, explicit_serialize, FWAction, Firework, Workflow
 from pymatgen.core import Structure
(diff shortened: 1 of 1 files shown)


atomate/vasp/firetasks/electrode_tasks.py
```diff
@@ -1,4 +1,5 @@
 import math
+from pydash import get
```
utf, line 2: I don't think you need this import?

utf (commented):
Thanks. Just one minor comment.

acrutt:
@utf do you have any thoughts on the matsci.org thread ([https://matsci.org/t/insertion-worflow-optimizefw-kwargs-issue/36593/6](https://matsci.org/t/insertion-worflow-optimizefw-kwargs-issue/36593/6))? I'm not fully sure whether setting NSW=299 is necessary. It would simplify debugging this to just remove that setting which seems unnecessarily high.

atomate/vasp/firetasks/electrode_tasks.py
acrutt: My mistake! I removed this import statement.

PR #675, opened 2021-06-01
Correct VASP input for elastic workflow

mjwen:
## Summary

Correct VASP input params that causes large error in elastic tensors:

- Explicit set `LREAL=False` to override  MPRelaxSet default (`Auto`). `False` is recommended by VASP manual for small systems and high accuracy. 
- Remove `PREC=High` for static calculations, which is inconsistent with the relaxation job setting: `Accurate`.


utf:
Fantastic thanks.

PR #720, opened 2021-09-13
Fix bug in VASP IO write when `spec_structure_key` defined

alchem0x2A:
## Summary
Current version of `WriteVaspFromIOSet` does not update the POSCAR using the `spec_structure_key` due to a bug on line 83 of `write_inputs.py`.

* Fix 1
`WriteVaspFromIOSet` does not correctly update the `vis` when `spec_structure_key` is set. The PR fixes the bug so the user-provided spec key can be passed.


itsshawnzhang:
Thanks for your contribution @alchem0x2A . We will merge this PR. Please let us know for any problems or concerns. Cheers!

PR #730, opened 2022-01-10
Fix CI

janosh:
Follow up to #724.

PR #734, opened 2022-01-13
Replace CircleCI with GitHub Action

janosh:
Continuing first attempt that was aborted in #730.


janosh:
After a lot of fiddling, `openbabel` finally successfully installs (`sudo ln -s /usr/include/openbabel3 /usr/local/include/openbabel3` [as suggested by @fredrikw](https://github.com/openbabel/openbabel/issues/2408#issuecomment-1014466193) did the trick) and `pytest` can actually run. For some reason, GH runs way longer than CircleCI (47 min vs 6 min). Not sure why. Doesn't look like CircleCI does any under the hood caching.

GH action:

```sh
= 2 failed, 116 passed, 82 skipped, 224 warnings, 4 errors in 2634.36s (0:43:54) =
```

CircleCI:

```sh
========== 124 passed, 80 skipped, 650 warnings in 202.02s (0:03:22) ===========
```

In any case, we might be able to bring this down a lot by using something [`pytest-split`](https://github.com/jerry-git/pytest-split).

janosh:
We need to pin `pymongo<4.0.0` because on CircleCI, deps were installed with `conda` whereas on with  GH action we're using `pip instal`. On Anaconda, the [latest `pymongo` is still v3.12.0](https://anaconda.org/anaconda/pymongo) whereas on PyPI, [it's already v4.0.1](https://pypi.org/project/pymongo/4.0.1).

`atomate` is not upwards compatible due to e.g.

```py
/home/runner/work/atomate/atomate/atomate/utils/testing.py:88: in tearDown
    for coll in db.collection_names():
TypeError: 'Collection' object is not callable. If you meant to call the 'collection_names' method on a 'Database' object it is failing because no such method exists.
```

janosh:
Looks like there were a few tests in

- `atomate/vasp/analysis/linear_response.py`
- `atomate/vasp/workflows/tests/test_insertion_workflow.py`

that would have failed even after the "Fix CI" PR #730 if they hadn't been skipped by `pytest`. Pinning `pymongo<4.0.0` or something else seems to have unskipped them run and now they fail. Will fix as part of this PR.

janosh:
All green! 🎉 🟢 

Just a final note that [my comment above](https://github.com/hackingmaterials/atomate/pull/734#issuecomment-1014645064) about GHA being slow was premature.

The much longer run time on GH action was due to me not spinning up a MongoDB server there yet. So every DB connection was timing out which of course took a long time. With that fixed, GH action is still a small amount slower (10m 24s circle vs 12m 25s GHA).

I asked @asottile about that who said

> I think their machines might be slightly beefier -- also circle tends to start sooner than GHA -- circle was basically the only one on par with http://pre-commit.ci in start time (not in execution time though1)

So this slight slowdown seems to be expected. Hence, ready for review here.

janosh:
One more point: I think we should keep the CircleCI workflow running for a while to make sure there are no further discrepancies between CI runs. Maybe after we transition to latest pymatgen with only py38+ support, we can drop CircleCI unless @computron or someone wants to keep it permanently.

computron:
Merging this prior to new atomate release

PR #738, opened 2022-01-19
Drop py37 support

### LINKED ISSUE #735: Drop py37
I think we should follow suit and drop Python 3.7 support so that `atomate` is compatible with the latest `pymatgen`.

janosh:
Closes #735.

Motivation: compatibility with latest `pymatgen`


janosh:
I think we can remove the `py3devtest` job as part of this PR as it's _exactly_ the same as `py3test` except it doesn't run on the `main` branch (not sure the what for) meaning it just burns compute for no added test coverage.

janosh:
Note that I removed the that `add_tags` and `add_filename` powerups from `atomate/vasp/powerups.py` that were deprecated in favor of identical functions in  `atomate/common/powerups.py` in https://github.com/hackingmaterials/atomate/commit/fe7ba87339343a9a13fdd1fc1c83fa9cd0373df8 almost a year ago. Seems like sufficient time for people to update their code. This makes this a PR a breaking change for people with py38+. People on older Python's are safe since this package will no longer be updated for them.

janosh:
CircleCI still fails on `Lobsterin._get_potcar_symbol()` because fix for `TypeError: a bytes-like object is required, not 'str'` in latest `pymatgen` was not released to `conda-forge` yet. GHA installs from PyPI. I guess we could retry pip install in Circle with latest `pymatgen` release or wait for update to appear on `conda`.

JaGeo:
also the MAGMOM fix is not yet in a released version.

JaGeo:
Thanks for your work. Yeah, it looks like the INCARs produced by the LobsterSet are now different than before?! Would need to inspect further...

JaGeo:
We can fix the error by adapting the test files a bit. I am on it. However, it shows me that I might need to work on the LobsterSet in pymatgen further to ensure that such changes do not happen that easily... 

I had to change 3 different test files. Now, everything should be working. You can find my changes here: https://github.com/jageo/atomate

janosh:
@JaGeo Excellent, thanks!

janosh:
The Lobster INCAR fix worked. All green here, ready for review.

JaGeo:
Great work @janosh  
Just to add here: I would be very happy about a new version of atomate including these changes.

Zhuoying:
@janosh @JaGeo Thanks for your contributions! It looks great (all green), so I am merging it for your convenience to use it. 
If there is anything wrt python3.7 drop later when running, please feel free to let us know. Thanks!

janosh:
@JaGeo I believe @computron would have to be the one to release a new version. Since this PR was a breaking change due to dropping 2 deprecated power ups, it would be a major version bump if `atomate` followed semantic versioning. Looking at passed versions this does not appear to be the case though, hence maintainers may wish to delay a new release instead.

PR #739, opened 2022-01-25
use pytest-split in GHA test workflow

janosh:
Following the initially longer than expected runtime of `pytest` in #734, I looked around a bit for a pytest plugin that would automatically distribute the workload among multiple concurrent runners.

Luckily, exactly such a thing exists. Never used it before but looks great.

https://github.com/jerry-git/pytest-split

And they even have a GHA demo:

https://github.com/jerry-git/pytest-split-gh-actions-demo/tree/master/.github/workflows

Let's see if I set this up correctly. :)

### DIFF
--- a/.github/workflows/combine_durations.py
+++ b/.github/workflows/combine_durations.py
@@ -1,0 +1,27 @@
+import json
+from pathlib import Path
+
+split_prefix = "split-"
+durations_path = Path(".test_durations")
+
+split_paths = Path(".").glob(f"{split_prefix}*/{durations_path.name}")
+
+try:
+    previous_durations = json.loads(durations_path.read_text())
+except FileNotFoundError:
+    previous_durations = {}
+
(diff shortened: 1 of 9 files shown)


itsshawnzhang:
Hello @janosh! Sorry about the delay. I will merge this PR now and please let me know for any other problems or concerns.
Thank you for the contribution. Cheers! :)

janosh:
Tbh, I'm not sure the potential CI speed up from this change is worth the added complexity. I guess we can try it out on `main` and revert if it causes any issues.

itsshawnzhang:
@janosh Sure thing!

itsshawnzhang:
@janosh I think I need to revert this PR since several errors arose by the `get_valid_path` in `drone.py` from `qchem` package.

janosh:
Feel free.

PR #745, opened 2022-02-02
Fix gzip contcar issue

fraricci:
This should fix the issue of the interpolation FW which currently does not find the CONTCAR file if its gzipped.
If the CONTCAR file is not gzipped, it still works fine.

More generally, this fix allows CopyFilesFromCalcLoc to decompress compressed file after copy if wanted.

This fix should solve previous raised issues (I can find them back now).


itsshawnzhang:
Thanks @fraricci for your contribution. I will merge this PR now and please let me know if there is any further problems or questions. Cheers! : )

PR #746, opened 2022-02-08
Fix PyPI page

janosh:
Follow up to #740.

Noticed that `atomate`'s PyPI page is still completely messed up because `setup.py` did not set

```py
long_description_content_type="text/markdown"
```


janosh:
Current state of PyPI page:

Screen Shot 2022-02-08 at 10 20 29

PR #749, opened 2022-03-01
small fixes to tests

jmmshn:
## Summary

- fixed redundant code in insertion test
- changed from `boto3.client` to `boto3.resources` to better match what `maggma` does.  Also I was failing tests because it seems like the AWS API no longer liked me using `boto3.client.create_bucket` in the way this was originally written.
- Install attempts produced an error saying that have to choose a different `scipy` version (Ubuntu with python 3.9 and 3.10) so I bumped it to the nearest > version.
- Not super sure about the last one: but since we have turned off `qchem` tests on the GitHub workflows should we turn it off in `setup.cfg` as well?  It took a long time for me to figure out why all my tests were failing.

### DIFF
--- a/atomate/utils/tests/test_database.py
+++ b/atomate/utils/tests/test_database.py
@@ -38,7 +38,7 @@ class DatabaseTests(unittest.TestCase):
 
     def test_s3_valid(self):
         with mock_s3():
-            conn = boto3.client("s3")
+            conn = boto3.resource('s3', region_name='us-east-1')
             conn.create_bucket(Bucket="test_bucket")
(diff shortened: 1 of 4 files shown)


Zhuoying:
Hi, @jmmshn Thanks for this PR. I will merge it for now. If there is an issue later, pls feel free to let us know.

BTW: The failed pytest actually was fixed already (https://github.com/hackingmaterials/atomate/pull/747). It would be great if you could **git pull** first before git push to avoid this for the next time. Please bear my OCD for all **green** tests.😂

PR #750, opened 2022-03-03
atwf add POSCAR not working

jmmshn:
`atwf add POSCAR` errors out due to the `yaml.load` command here

The current version of `yaml.load` requires a `Loader` kwarg or just calling `safe_load` instead.
See here for more detail.
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

### DIFF
--- a/scripts/atwf
+++ b/scripts/atwf
@@ -78,7 +78,7 @@ def _get_wf(args, structure):
         return func(structure)
 
     else:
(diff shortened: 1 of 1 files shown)


itsshawnzhang:
I have reviewed the file changed and I will merge this PR now. Thanks for your contribution and please let us know if there is any further problems or concerns. Cheers! :)

PR #764, opened 2022-09-20
Remove repeated words

janosh:
@Zhuoying Just ran a regex over the code base that removed repeated words.

CI is still failing atm but these are safe changes. Also Jimmy told me he fixed errors locally and will submit a PR for it soon.


Zhuoying:
Thanks for the clean-up

PR #768, opened 2023-01-09
Fix for MagneticOrderingsWF static only

fraricci:
I realized that the MagneticOrderingsWF did not work properly when static = True (static calculation only).
Here it is a fix that seems to work fine.

To note that I had to code explicitly "structure is not None" in the StaticFW.
I'm not sure why, probably it's due to the fact that the structure generated
in a MagneticOrderingsWF is a TransformedStructure obj and not a standard Structure obj.

I fixed also a bug in the calculation of the total magnetization per formula unit.


Zhuoying (commented):
Thanks for reporting this bug. The term `num_formula_units` needs to be corrected rather than the way to calculate `total_magnetization_per_formula_unit`. Could you do a quick fix and I will merge this PR after all tests pass.

fraricci:
Let me know if anything else is needed here.
I think the checks are failing for reasons unrelated to my commits.
Thanks

Zhuoying:
Hi @fraricci, thanks for the fix. 
I recently merged some PRs to fix the previous circleci test failures (they should not be related to your PR). 
Could you git pull these changes and retrigger the test? If the PR passes the pytest, I am merging it right now. 
Thanks!

PR #786, opened 2023-12-05
Add hashes to Q-Chem task documents

espottesmith:
The (relatively) new Materials Project molecules database ([here](https://next-gen.materialsproject.org/molecules)) uses Weisfeiler-Lehman graph hashes, primarily as an aspect of molecule IDs. Currently, to add tasks to our build pipeline, I have a script to incorporate the hashes into completed Q-Chem task documents. This PR instead directly incorporates the hashes in task document construction at the time of drone assimilation.

Hashes have been incorporated in the Q-Chem drone tests, and these tests pass locally.


espottesmith:
So several tasks _not_ related to graph hashes and Q-Chem docs failed. @samblau I think one of them is related to the fragmenter. Any ideas?

samblau:
Honestly, we should probably just deprecate / remove the fragmenter. In practice, it basically always made sense to do fragmentation in a script to customize it to a given application's species generation. I don't think the fragmentation workflow in Atomate really has any value, so if it's breaking tests, let's just get rid of it. 

More broadly, I've stopped using Atomate in favor of QuAcc (and others have probably stopped using it in favor of Atomate2). I think Rishabh was working on Q-Chem stuff in Atomate2, which I think also included a new task doc schema? So it might be worth working in these changes there too.

